PR #4, opened 2015-09-08
Expose HTTP server to support calling `close`

dhm116:
In order to support cleanly shutting down the express app after calling `run`, the result of calling express's `listen` method must be returned. This adds support for:

```javascript
var server = s3rver
      .setHostname('localhost')
      .setPort(4568)
      .setDirectory(S3_DIR)
      .setSilent(true)
      .run(function (err, host, port) {

      });

server.close(function(err) {

});
```

This is a pretty essential feature during testing.


jamhall:
Thank you!

Issue #6, opened 2015-09-18
List buckets


Lochemage:
Hi, I really like that you have taken fakes3 and ported it to Node.js, since my main project is already in Node it keeps me from setting up Ruby etc.

Is there any chance for implementing list bucket operations for this?

jamhall:
Hi there! What do you mean by list buckets? Is this an amazon s3 feature? Do you want a command or a URL to list the buckets? Please give me more info :) 

One of the main reasons I ported it to nodejs is because fakes3 objects weren't persistent, i.e. every time you restarted the server, you lost access to the objects! :(

Lochemage:
@jamhall 

Yes, I'm talking about s3's ability to retrieve a listing of all buckets or object keys within a bucket.

http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listBuckets-property
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjects-property

Also, I believe this is not the case any longer with fakes3. At least I am able to list and fetch objects that I put there from before a restart.

jamhall:
This functionality definitely exists in s3rver. Please have a look in the test directory for examples :)

Lochemage:
@jamhall 

Hmm, then I will have to investigate further, because all of my unit tests that test bucket listing all fail with a result size of 0.  It may just be that the data I get is not aws sdk compatible.

jamhall:
What happens when you go to the bucket url directly? http://localhost:yourChosenPort/yourBucketName . That should return an Xml list of bucket objects.

Lochemage:
@jamhall 

Okay, after some investigating it seems like I am getting a full listing from the `Contents` list only if I use an empty Prefix value. Any prefix value causes the result list to come up empty.

Just to double check, I have run the unit tests on your repo and those same tests fail as well.

This may be related to me using Node.js version 4.0.0, I know there have been a lot of changes in it.

Lochemage:
Here is the output I am getting.

```
  S3rver Tests
    √ should fetch fetch five buckets
    √ should fail to create a bucket because of invalid name
    √ should fail to create a bucket because name is too long
    √ should fail to create a bucket because name is too short
    √ should delete a bucket
    √ should not fetch the deleted bucket
    √ should list no objects for a bucket
    √ should store a text object in a bucket
    √ should store a text object with some custom metadata
    √ should return a text object with some custom metadata
[...]

PR #8, opened 2015-09-23
listObjects compatibility with Windows OS

Lochemage:
This is a fix for issue #6.

On the Windows OS, file paths use the backslash to delimit each folder while object keys expect the forward slash. When walking file storage items, we explicitly convert the slashes so they match our expected `object key` format.

I have run this change through this projects unit tests as well as my own and they all pass now.


jamhall:
Thanks very much! :)

Issue #9, opened 2015-09-23
listObjects delimiter not implemented?


Lochemage:
My projects' unit tests fail when trying to list objects using a delimiter value (namely to find objects contained within a folder but not including objects within sub-folders).

As far as I can tell, this feature is only partially implemented. I am able to supply a delimiter option, but it is not actually used.

PR #11, opened 2015-09-23
Added support for delimiter option when listing objects.

Lochemage:
Also added unit test for it.


Lochemage:
Oh, this is in reference to issue #9

jamhall:
Many thanks for your contributions @Lochemage !

Issue #10, opened 2015-09-23
objects.CommonPrefixes


Lochemage:
This is part of the object structure retrieved by `listObjects()`, similar to `objects.Contents`, which lists out all keys that match between the `prefix` and the `delimiter`. Note, this value only exists if a delimiter is used.

The definition for the `CommonPrefixes` property is:
**Gets the CommonPrefixes property. A response can contain CommonPrefixes only if you specify a delimiter. When you do, CommonPrefixes contains all (if there are any) keys between Prefix and the next occurrence of the string specified by delimiter. In effect, CommonPrefixes lists keys that act like subdirectories in the directory specified by Prefix. For example, if prefix is notes/ and delimiter is a slash (/), in notes/summer/july, the common prefix is notes/summer/.**

Source: http://docs.aws.amazon.com/sdkfornet1/latest/apidocs/html/T_Amazon_S3_Model_ListObjectsResponse.htm

This issue relies on issue #9 to support delimiters.

jamhall:
Thanks!

Lochemage:
Works now!

PR #12, opened 2015-09-23
CommonPrefixes support

Lochemage:
Fixes issue #10 

This expands the previous support for using Delimiters and also includes the CommonPrefixes data that is supposed to be returned by the server.


jamhall:
Thank you!

PR #13, opened 2015-10-23
Fix program option --indexDocument

peterkuma:
Fix typo and missing option value.

Not sure if `[value]` or `<value>` would be preferred (?).


leontastic:
This was fixed in a8170ed69922f3e12cdcdf15014b0d23cfa9145f.

PR #14, opened 2015-10-23
Fix ETag field format in HTTP response

peterkuma:
The ETag should be surrounded by double quotes.

Some client libraries such as boto require this format and refuse GET requests otherwise:

`IOError: Cannot store S3 object "bucket/object": ETag from S3 did not match computed MD5. 34913be17f33fe36466f5b932aff626f vs. 34913be17f33fe36466f5b932aff626f`

RFC: https://tools.ietf.org/html/rfc7232#section-2.3

Amazon also lists quoted ETags in examples in their documentation:

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

There were quite a few occurrences in the code - hopefully I found all of them. The tests are passing and boto works as expected.

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -15,7 +15,7 @@ module.exports = function (rootDirectory, logger, indexDocument, errorDocument) 
 
   var buildResponse = function (req, res, status, object, data) {
     res.header('Access-Control-Allow-Origin', '*');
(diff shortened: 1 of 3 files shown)


jamhall:
Interesting! Thank you for the info and the pull-request. Much appreciated!

PR #18, opened 2015-11-01
Allow CORS Requests

toehio:
This is a simple solution to allow all CORS requests, which are necessary when using s3rver for testing web applications. I tested it with Chromium.


jamhall:
Hello. It would be nice if CORs is activated via an option - it should be disactivated by default. What do you think?

toehio:
That's a good idea. I just added the option.

jamhall:
Excellent! Thanks very much. I'll push a new release to NPM tomorrow morning :+1:

jamhall:
@toehio  - v.0.0.10 has been [released](https://github.com/jamhall/s3rver/releases/tag/v0.0.10). It has also been pushed to NPM. Cheers

PR #25, opened 2016-01-16
Allows users to plug in other file systems providers instead of fs-extra only 

anitahitouch:
HI 
With these changes we can use any file system to be used as backend for s3rver .
Currently s3rver uses fs-extra as a file system for this server .  if a user has a file system which exposes functions with same semantics and signatures as that of fs file system but works under different environment or protocols  then we can use that in place of fs or fs extra

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -4,7 +4,7 @@ var app = function (options) {
       app = express(),
       logger = require('./logger')(options.silent),
       Controllers = require('./controllers'),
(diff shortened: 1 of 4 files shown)


jamhall:
Hi there! Interesting idea! Considering s3rver is to be used for development purposes, I'd be surprised if someone uses a different type of file system to store their objects, but I like the idea. Thank you for your contribution. I have a couple of extra code changes to do before pushing a new release to NPM. I'll make a new release, probably on Monday, if not on Tuesday at the latest.

PR #29, opened 2016-01-28
Log errors if cannot listen (most often [Error: listen EADDRINUSE 127…

antono:
….0.0.1:4568])

Usee process.exit() to exit


jamhall:
Thank you!

Issue #30, opened 2016-02-16
Object Copy does not work


Sparragus:
Sorry for the click-bait issue name. Object Copy does work, but read on and see what I mean.

I've been having a discussion over at https://github.com/aws/aws-sdk-js/issues/901 regarding their documentation on `s3.copyObject(...)`. In the same document where `x-amz-copy-source` is specified, there are conflicting descriptions on how it should be set.

### AWS Documentation

The 'Syntax' section of the [documentation](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html#RESTObjectCOPY-requests) says: 

```
x-amz-copy-source: /source_bucket/sourceObject
```

However, in the 'Request Headers' section, the [documentation](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html#RESTObjectCOPY-requests) says:

```
Name:
   x-amz-copy-source
Description:
   The name of the source bucket and key name of the source object, separated by a slash (/).
  
   Type: String

   Default: None
Required:
   Yes
```
So, which one should be followed?
   1. '/' + bucketName + '/' + keyName
   2. bucketName + '/' + keyName

### What s3rver does
`s3rver` expects `x-amz-copy-source: /source_bucket/sourceObject`. See [lib/controllers.js#L264-L268](https://github.com/jamhall/s3rver/blob/master/lib/controllers.js#L264-L268).

Without the initial '/', `srcBucket = srcObjectParams[1]` ends up being the source object's `key`. Eventually this will fail with a
```
error: No bucket found for "image.png"
info: PUT /BUCKET_NAME/image.copy.png 404 207 - 60.642 ms
```

### What should we do?
I think that in the mean time we should do something to accept both forms. Otherwise integration with other libraries fail. `s3fs` is an example. See [s3fs/lib/s3fs.js#L345](https://github.com/RiptideElements/s3fs/blob/a39b0b716e96190fff620e793e92cafbe6a90c7f/lib/s3fs.js#L345) (It does not use an initial '/')

However, I invite you all to participate on the discussion over at https://github.com/aws/aws-sdk-js/issues/901.

PR #31, opened 2016-02-16
Normalize x-amz-copy-source

Sparragus:
Fixes https://github.com/jamhall/s3rver/issues/30

Since S3's documentation is ambiguous, `s3rver` should expect both versions of `x-amz-copy-source`. This change adds a '/' to `copy` in case `x-amz-copy-source` does not have a leading '/'.

For S3's documentation, see https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html


jamhall:
Hello @Sparragus ! Thank you for reporting this problem!. I'll update NPM.

PR #40, opened 2016-05-25
Support setting the ContentDisposition parameter, other general improvements

rockmacaca:
Sorry these commits are a bit all over the place, there were just a few features I needed immediately and didn't organize them into separate branches for separate pull requests. Let me though if this is problematic.

I'll just outline in a bit more detail exactly what these do:

1. Instead of returning an `app` object with just the custom serve method, I extended the express app with the serve method and returned the express app. This allows s3rver to run on the same port as an existing web server when using express's app mounting feature.
2. There was a bug that prevented using prefix matching and possibly some other path-related functions on Windows since the `fs` module adapts to the OS when creating path strings. I just used `url.format()` to essentially replace `\\` with `/` in the paths, functionality now appears to be restored.
3. When you pass the necessary fields in the options object for s3rver, `app.serve()` will now attempt to listen on HTTPS. I put a note in the readme for setting up the AWS SDK to work with s3rver running in HTTPS mode.
4. Store and set the ContentDisposition header if provided. The main use of this is for setting the filename of a download and whether it should be downloaded or attempted to be displayed inline. I've tested code utilizing this header on s3rver and on the real thing, this part definitely works.

I've been actively using my fork and I think these changes are worthy of being merged back in, again let me know if there's anything I need to adjust here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ s3rver --help
 ## Supported clients
 
 Please see [Fake S3s wiki page](https://github.com/jubos/fake-s3/wiki/Supported-Clients) for a list of supported clients.
+When listening on HTTPS with a self-signed certificate, the AWS SDK in a Node.js environment will need ```httpOptions: { agent: new https.Agent({ rejectUnauthorized: false }) }``` in order to allow interaction.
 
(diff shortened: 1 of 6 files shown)


jamhall:
This is excellent. Thank you!

PR #41, opened 2016-06-21
Audio/Video files not served properly - unable to scrub with HTML5 player

rockmacaca:
This might be an easy fix, but I don't have much experience messing around with the boilerplate of an Express app to get large media files served correctly.

Imgur

Actual S3 response:

Imgur

I'll be looking into this and will submit a pull request if I figure something out.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -6,7 +6,6 @@ module.exports = function (options) {
       logger = require('./logger')(options.silent),
       Controllers = require('./controllers'),
       controllers = new Controllers(options.directory, logger, options.indexDocument, options.errorDocument, options.fs),
-      concat = require('concat-stream'),
       path = require('path'),
       https = require('https');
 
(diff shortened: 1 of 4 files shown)


rockmacaca:
So this pull isn't final, still need to fix a few things. I can't seem to figure out why Chrome refuses to play the content with my implementation so far. The partial requests seem to look fine in Edge and in VLC.

rockmacaca:
This ended up being not as hard as I thought it would be, the only deficiencies with this implementation I can think of is the lack of support for `If-Range`, the server will not respond with `416 Requested Range not satisfiable`, and index documents for static web hosting won't respect the `Cache` header.

If we were to give these proper support, I think it would be best to take advantage of the [send](https://github.com/pillarjs/send) library, which basically takes care of most of the logic of reading from the filesystem, detecting/generating metadata such as `Content-Length` and `ETag` and sending the response.

RonanDrouglazet:
👍 fix my issue on delivering mediafile for safari (mobile or desktop) because req Content-Range was not supported since your fix ! I hope this will validated and merged :)

jamhall:
Hello! Thank you very much @rockmacaca  for your contributions.

The first stable release, v1.0.0 has been released onto npm.

Thanks again

RonanDrouglazet:
thanks @jamhall !

rockmacaca:
No problem, glad I could help getting to 1.0!

PR #42, opened 2016-06-26
Files likely not closed after writing

rockmacaca:
This was probably introduced in eeb5793 after I changed `fs` calls to use streams. On Windows, Node.js still has a filesystem lock on new files and attempting to delete (and likely overwrite) them will result in
```
EPERM: operation not permitted, stat 'c:\...'
```
until the offending process is restarted.

I think I know where to look to fix it, I should have time tomorrow to attach a pull request if this isn't resolved by then.

### DIFF
--- a/lib/file-store.js
+++ b/lib/file-store.js
@@ -297,6 +297,7 @@ var FileStore = function (rootDirectory,fs) {
       return done('Error writing file');
     });
     writeStream.on('finish', function () {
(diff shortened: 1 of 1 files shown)


jamhall:
Thanks :) I'll push to npm tomorrow.

jamhall:
v1.0.1 has been published onto NPM

PR #44, opened 2016-07-03
Support domain-style bucket names.

LegNeato:
The regex was wrong. In order to support domain-style buckets with
subdomains, `.` must be allowed. This now better matches the S3 docs.


jamhall:
Awesome! Thanks!

jamhall:
v1.0.2 has been released onto NPM. :)

LegNeato:
Thanks for the quick turnaround!

PR #46, opened 2016-08-19
Add Log Label

peterschwarz:
Add a log label to create a clear delineation between S3rver logs and an users application.  This is particularly useful when using S3rver programmatically within an application (e.g. during development or testing time).


jamhall:
Thank you!

PR #49, opened 2016-10-20
Respond to requests on bucket vhost

jbergknoff:
I have a situation where I'm testing S3 interaction code as a black box. I can't (or prefer not to) modify the code to point to a special S3 endpoint and force path-style requests. Instead, what I'd like to do is intercept traffic bound for `[bucket].s3.amazonaws.com` and redirect that traffic to s3rver, e.g. with a rule in /etc/hosts.

This PR has s3rver parse the `host` header and, if it matches a [bucket vhost](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro), then it handles the request for that bucket.

This only changes the behavior when s3rver receives a request with a host like `[bucket].s3(-[region])?.amazonaws.com`, so normal usage stays the same.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -23,7 +23,11 @@ module.exports = function (options) {
   app.use(function (req, res, next) {
     var host = req.headers.host.split(':')[0];
 
-    if (options.indexDocument && host !== 'localhost' && host !== '127.0.0.1') {
+    // Handle requests for <bucket>.s3(-<region>?).amazonaws.com, if they arrive.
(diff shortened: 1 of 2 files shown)


slobo:
This would be useful to us esp if we were somehow able to configure the custom domain to match (`[bucket].s3rver.our.domain`), maybe using `env` or some cli argument...

jbergknoff:
@leontastic this PR is related to #50. I've rebased this onto master as well now.

leontastic:
@jbergknoff Clever solution to this problem. Thanks for your contribution!

Issue #51, opened 2016-12-15
Incorrect Error Message


aniltallam:
For 'NoSuchBucket' error code, currently 'The resource you requested does not exist' is returned. But the correct message is 'The specified bucket does not exist'.

Ref: http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

aniltallam:
Sent you pull request 'Fixe for #51: Incorrect Error Message '. 
Pls accept it.

PR #52, opened 2016-12-15
Fixe for #51: Incorrect Error Message

aniltallam:
Changed Error Message for 'NoSuchBucket' to 'The specified bucket does not exist'


jamhall:
Thank you!

PR #54, opened 2017-01-25
Use fs provided in options for all file operations

laurynasl-wix:
Currently, not all file operations use the provided fs. I've updated also the places where there  were such operations. This makes it possible to use the module with something like https://github.com/webpack/memory-fs


jamhall:
Thank you!

PR #56, opened 2017-02-23
Preserve Content-Disposition in copyObject, fix If-None-Match header

rockmacaca:
Just realized I've been holding on to these changes for a while now, I guess I hadn't realized I wasn't using your npm packages.


jamhall:
Awesome! Thank you!

rockmacaca:
Wow, that was quick!

PR #57, opened 2017-02-23
Add support for copying objects with metadata replace

MartinSoto:
Hi, thanks a lot for S3rver!

Our S3-based application uses object copying with metadata replacement, and we realized that it's not supported by S3rver. I just added the missing support. The relevant S3 operation is [PUT Object - Copy](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html), in particular, the `x-amz-metadata-directive` header.


jamhall:
Hi @MartinSoto  - thank you for your contribution!

PR #59, opened 2017-03-05
Expose the Express app on S3rver instances

specialkk:
The fs option added in #54 wasn't documented and was a breaking change, so I was pretty confused what was wrong for a bit. Using `fs-extra` by default makes it non breaking : ).


rockmacaca:
@jamhall Any comment on this?

rockmacaca:
OK, I've figured out a better solution since the problem was just that certain default options aren't set when you try to use the Express app via `require('s3rver/lib/app')`. Instead, I've opted to make it accessible so something like `new S3rver(options).app` is possible. The `app` prop just gets recreated every time the options are updated.

specialkk:
Closing since I'll include these changes in a less intrusive manner in #63.

PR #61, opened 2017-03-10
return Last-modified and Etag headers with 304 repsonse

byteio:
Hello! 
In its current state, the server does not return cache headers when returning a 304 response. I've submitted this PR for your review to resolve this issue. 

The current behavior is causing some of my tests to fail in a project I'm working on because my code relies on the production AWS behavior which is to return these headers with the 304 response.

It's a pretty simple change, but please let me know if I can adjust the code in any way to better accommodate the project's preferred style. Thanks!

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -19,8 +19,6 @@ module.exports = function (rootDirectory, logger, indexDocument, errorDocument, 
 
   var buildResponse = function (req, res, status, object, data) {
     res.header('Access-Control-Allow-Origin', '*');
-    res.header('Etag', '"' + object.md5 + '"');
-    res.header('Last-Modified', new Date(object.modifiedDate).toUTCString());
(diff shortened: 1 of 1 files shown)


lib/controllers.js
```diff
@@ -19,8 +19,6 @@ module.exports = function (rootDirectory, logger, indexDocument, errorDocument,
 
   var buildResponse = function (req, res, status, object, data) {
     res.header('Access-Control-Allow-Origin', '*');
-    res.header('Etag', '"' + object.md5 + '"');
-    res.header('Last-Modified', new Date(object.modifiedDate).toUTCString());
```
leontastic: Does your addition work without removing these lines? Remove these lines will cause `Etag` and `Last-Modified` values to disappear in other calls to  `buildResponse`.

PR #64, opened 2017-04-05
Do not fail when deleting a nonexistent object

vmaxv:
If the object specified in the request is not found, `s3rver` responds with 404 error, unlike Amazon S3 that returns the result as deleted for both [single](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html) and [multi-object](http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html) operations.

Align `deleteObject` and `deleteObjects` with the API spec and return a success response even when the target object does not exist.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -456,6 +456,10 @@ describe('S3rver Tests', function () {
     });
   });
 
+  it('should not fail to delete a nonexistent object from a bucket', function (done) {
+    s3Client.deleteObject({Bucket: buckets[0], Key: 'doesnotexist'}, done);
+  });
+
   it('should fail to delete a bucket because it is not empty', function (done) {
     s3Client.deleteBucket({Bucket: buckets[0]}, function (err) {
       err.code.should.equal('BucketNotEmpty');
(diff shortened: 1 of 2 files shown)


vmaxv:
@jamhall Did you have a chance to look at it?

test/test.js
```diff
@@ -227,7 +227,7 @@ describe('S3rver Tests', function () {
         if (err) {
           return done(err);
         }
-        object.ContentLength.should.equal(stats.size.toString());
+        object.ContentLength.should.equal(stats.size);
```
leontastic: Can you remove this changes and rebase on `master`? This was fixed in #79

test/test.js
```diff
@@ -349,7 +349,7 @@ describe('S3rver Tests', function () {
           return done(err);
         }
         object.ETag.should.equal('"' + md5(data) + '"');
-        object.ContentLength.should.equal(data.length.toString());
+        object.ContentLength.should.equal(data.length);
```
leontastic: Can you remove these changes and rebase on `master`? This was fixed in #79

test/test.js
```diff
@@ -364,7 +364,7 @@ describe('S3rver Tests', function () {
...
```
leontastic: Can you remove these changes and rebase on `master`? This was fixed in #79

test/test.js
```diff
@@ -481,7 +485,7 @@ describe('S3rver Tests', function () {
         return done(err);
       }
       object.ETag.should.equal('"' + md5('Hello!') + '"');
-      object.ContentLength.should.equal('6');
+      object.ContentLength.should.equal(6);
```
leontastic: Can you remove these changes and rebase on `master`? This was fixed in #79

leontastic:
Another issue supporting the behaviour reported by @vmaxv: https://github.com/andrewrk/node-s3-client/issues/26

PR #67, opened 2017-05-10
add bucketName to getBucket response to match current AWS response

AlixBa:
If it's possible to bump a version with this modification, it would be perfect :D


KOnaitis:
+

leontastic:
@AlixBa

PR #73, opened 2017-06-19
S3 event notification for Put, Copy and Delete objects

anantab:


anantab:
I think the s3 event notification is a major feature missing from this library. I have added a feature where you can subscribe to a Observable for Put, Copy and Delete object events in buckets. The observable returns a JSON object similar to what AWS s3 event notification returns. I am relatively new in NodeJs, so any suggestions and improvements are more than welcome.

leontastic:
@anantab This looks like a really good feature for us to have. Would you mind rebasing on top of `master`, fixing any merge conflicts and re-testing the feature? Really appreciate the work you put into this, including clear docs for this feature.

PR #79, opened 2018-01-18
ci(node): Update runtime versions

n1ru4l:
cc: @leontastic

~~Some tests are failing. I am going to fix them.~~

Ready for merge ☺️

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -21,7 +21,7 @@ module.exports = function (rootDirectory, logger, indexDocument, errorDocument, 
     res.header('Access-Control-Allow-Origin', '*');
     res.header('Etag', '"' + object.md5 + '"');
     res.header('Last-Modified', new Date(object.modifiedDate).toUTCString());
-    res.header('Content-Type', object.contentType);
+    res.setHeader('Content-Type', object.contentType);
 
     if (object.contentEncoding)
       res.header('Content-Encoding', object.contentEncoding);
--- a/test/test.js
+++ b/test/test.js
@@ -227,7 +227,7 @@ describe('S3rver Tests', function () {
         if (err) {
           return done(err);
         }
-        object.ContentLength.should.equal(stats.size.toString());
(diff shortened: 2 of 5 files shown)


lib/controllers.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (rootDirectory, logger, indexDocument, errorDocument,
     res.header('Access-Control-Allow-Origin', '*');
     res.header('Etag', '"' + object.md5 + '"');
     res.header('Last-Modified', new Date(object.modifiedDate).toUTCString());
-    res.header('Content-Type', object.contentType);
+    res.setHeader('Content-Type', object.contentType);
```
n1ru4l: See: https://github.com/expressjs/express/issues/2697

AWS does not append the charset to the content-type.

test/test.js
```diff
@@ -227,7 +227,7 @@ describe('S3rver Tests', function () {
         if (err) {
           return done(err);
         }
-        object.ContentLength.should.equal(stats.size.toString());
+        object.ContentLength.should.equal(stats.size);
```
n1ru4l: see: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property

The examples show numbers instead of strings

test/test.js
```diff
@@ -793,7 +793,7 @@ describe('S3rver Tests with Static Web Hosting', function () {
         return done(new Error('Invalid status: ' + response.statusCode));
       }
 
-      if (response.headers['content-type'] !== 'text/html; charset=utf-8') {
+      if (response.headers['content-type'] !== 'text/html') {
```
n1ru4l: The previous content-type was wrong.
I created a bucket myself and enabled static web hosting. This is the correct content-type you should receive.

leontastic:
@n1ru4l Thanks! I noticed these tests failing too. I actually don't have write access just yet, pinging @jamhall to grant permission.

PR #96, opened 2018-01-19
Make tests independent

n1ru4l:
Currently the tests affect each other.
IMHO a test should be self contained.
Since the tests take some time, it can be really painful to debug a single test when they all affect each other.

@leontastic Are you okay with me using the promise instead of the callback api of the aws-sdk?

**TODO**
- [x] Fix failing tests


leontastic:
👍 100% agree about isolating tests given how large the test suite is.

I'd love to see all callbacks converted into promises, in the tests and source code.

n1ru4l:
@leontastic Lets merge this and do the test cleanup in a followup pull request

PR #78, opened 2018-01-16
Add support for multipart uploads

leontastic:
Resolves #33 

Multipart uploads is a major feature missing from this mock S3 server. Without it, it's impossible to test uploading files larger than 5MB using `aws-sdk`:

```javascript
const s3 = new AWS.S3()
// ...
s3.upload(params, (err, data) => {})
```

PR #98, opened 2018-01-22
Cleanup Option

### LINKED ISSUE #58: Cleaning up the directory after s3rver.close()
I am using s3rver in my mocha tests, it's great so far!

I have one recommendation, though:
When calling `s3Instance.close()`, the directory passed to the constructor should be cleaned up and emptied. It's more convenient that way;)

n1ru4l:
- Adds failing tests for new option `removeBucketsOnClose` (defaults to false).

When the option is set to true all buckets will be deleted when calling `server.close`.

closes #58

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -25,9 +25,9 @@ exports.walk = function (dir, fs) {
 function rmdir (fs, directory, callback) {
   fs.readdir(directory, function (err, list) {
     if (err) return callback(err);
-    async.everySeries(list, function (item, callback) {
+    async.eachSeries(list, function (item, callback) {
       var filename = path.join(directory, item);
-      var stats = fs.stat(filename, function (err, stat) {
+      fs.stat(filename, function (err, stat) {
         if (err) return callback(err);
         if (filename === '.' || filename === '..') return callback();
         if (stat.isDirectory()) return rmdir(fs, filename, callback);
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,4 +1,7 @@
 'use strict';
+var path = require('path');
+var async = require('async')
+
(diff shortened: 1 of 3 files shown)


n1ru4l:
Lol seems like the cleanup is already happening 😕

lib/utils.js
```diff
@@ -17,3 +20,23 @@ exports.walk = function (dir, fs) {
 
   return results;
 };
+
+// remove a path recursively (async)
+function rmdir (fs, directory, callback) {
```
n1ru4l: I used this over fs-extra emptyDir because it is possible to set fs as a aoption and other fs implementation do not have those methods.

leontastic: @n1ru4l Do you think we should remove the `fs` option? I can't think of a reason to continue including it as an option. We make use of a few `fs-extra`-only methods anyway.

n1ru4l: I can not find any issue related to why this option exists. The only benefit I could think of is a in-memory-fs or something similar.

Maybe @jamhall can explain why it is there 😅

leontastic: I would rather see `fs-extra` used directly. Would make things much simpler to call `fsExtra.remove` than rolling out a homemade implementation.

I don't have an opinion on removing the `fs` option, but if we do we can cut a major release for it.

This is my only gripe with this PR – otherwise I want this merged so we can cut a minor release with this (and multipart uploads #78) soon. @n1ru4l let me know what you decide, we can always refactor in the next release.

n1ru4l: Let's keep it like this for now and release a new minor version and proceed further when @jamhall has answered.

n1ru4l: @leontastic Just realized the next release must be ma major release anyway since we dropped support for node versions prior to 6.

Issue #101, opened 2018-01-24
MaxListenersExceededWarning when running the tests


n1ru4l:
```
  S3rver Tests
    ✓ should fetch fetch six buckets
    ✓ should create a bucket with valid domain-style name
    ✓ should fail to create a bucket because of invalid name
    ✓ should fail to create a bucket because of invalid domain-style name
    ✓ should fail to create a bucket because name is too long
(node:29986) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added. Use emitter.setMaxListeners() to increase limit
```

Introduced by #96. Possible memory leak?

n1ru4l:
The culprit is this line of code https://github.com/winstonjs/winston/blob/bd06c4091e9e248e6fed6278e0699e5cfdc7de24/lib/winston/exception-handler.js#L59

PR #104, opened 2018-01-24
Remove fs option and setters

n1ru4l:

PR #105, opened 2018-01-24
Release v2.0.0

leontastic:


leontastic:
Hoping to draft a release to make multipart uploads available #78 and remove the `fs` option from initialization #104. @n1ru4l Is there anything else you think we should wait for before the release? If not I will merge, tag and publish to npm.

leontastic:
Actually, I think we should probably fix #101 before doing the release.

n1ru4l:
@leontastic I started creating a release draft with a changelog here: https://github.com/jamhall/s3rver/releases

Would be nice if you can complete/review it! I will also investigate on the issue.

PR #108, opened 2018-01-24
Change var to const/let

n1ru4l:
I prefer const/let over var 😅


specialkk:
If we're targeting Node 6, you can also remove all the `'use strict';`  statements since it's automatic for modules :P

specialkk:
Whoops I forgot that's not true for CJS modules, I'm too used to Babel.

PR #114, opened 2018-02-01
S3 event notification for Put, Copy, Post and Delete objects

anantab:
Rebased on master branch


lib/app.js
```diff
@@ -6,9 +6,12 @@ const createLogger = require('./logger');
 const path = require('path');
 const https = require('https');
 const morgan = require('morgan');
+const subject = require('rxjs/Subject').Subject;
+require('rxjs/add/operator/filter');
 
 module.exports = function (options) {
   const app = express();
+  app.s3Event = new subject();
```
n1ru4l: If it must be an constructor please capitalize `subject`

leontastic:
@anantab Can you please rebase this branch off `master` again? Sorry for deferring this PR behind others. Once you clean up this branch I'll perform another review and try to get this into the next minor release.

anantab:
@leontastic Rebased.

bin/s3rver.js
```diff
@@ -63,3 +63,17 @@ new S3rver(program).run(function(err, host, port) {
     console.log("now listening on host %s and port %d", host, port);
   }
 });
+
+s3rver.s3Event.subscribe(function(event) {
+  // eslint-disable-next-line no-console
+  console.log(event);
```
leontastic: @anantab can we disable this logging based on the `silent` CLI option? Would prefer to make logging configurable instead of forcing users to see noisy logs.

Alternatively you could add another CLI option to turn on/off S3 event logging independently of the internal logger.

bin/s3rver.js
```diff
@@ -63,3 +63,17 @@ new S3rver(program).run(function(err, host, port) {
...
+});
+
+s3rver.s3Event
+  .filter(function(event) {
+    return event.Records[0].eventName == "ObjectCreated:Copy";
+  })
+  .subscribe(function(event) {
+    // eslint-disable-next-line no-console
+    console.log(event);
```
leontastic: Same comment as above – logs should be configurable, so we don't force users to accept noisy logs.

lib/app.js
```diff
@@ -7,9 +7,12 @@ const path = require("path");
 const Controllers = require("./controllers");
 const cors = require("./cors");
 const createLogger = require("./logger");
+const subject = require("rxjs/Subject").Subject;
+require("rxjs/add/operator/filter");
 
 module.exports = function(options) {
   const app = express();
+  app.s3Event = new subject();
```
leontastic: Please capitalize classes instantiated with `new`:
```node
new Subject()
```

lib/file-store.js
```diff
@@ -380,6 +381,7 @@ const FileStore = function(rootDirectory) {
           if (err) {
             return done("Error updating metadata");
           }
+          metaData.key = destKey;
```
leontastic: @anantab Can you move this into the `createMetaData` call? It doesn't look like `key` is being used in `createMetaData`, so you can change `key: srcKey` to `key: destKey` instead to achieve the same result. This way we don't increase coupling between these callbacks and `createMetaData`, and you can delete line 339.

lib/models/s3-event.js
```diff
@@ -0,0 +1,83 @@
...
+  const event = {
+    "Records": [
+      {
+        "eventVersion": "2.0",
+        "eventSource": "aws:s3",
+        "awsRegion": reqHeaders.host,
+        "eventTime": eventData.S3Item.creationDate || new Date().toISOString(),
+        "eventName": eventName,
+        "userIdentity": {
+          "principalId": 'AWS:' + principal[1],
```
leontastic: This doesn't match the format of the example given in the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html). @anantab do you observe this value prepended with `AWS:` when you get events from real S3?

lib/models/s3-event.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+const S3Event = function (eventData, reqHeaders) {
+  var principal = [];
+  try {
```
leontastic: Instead of using a `try..catch` block and retrieving an element from the `principal` array later in this function, can you use an `if..else` block or ternary and assign a precise `principalId` value here? The current way this is written increases the complexity of understanding your intent.

README.md
```diff
@@ -66,6 +67,36 @@ mysite.local 127.0.0.1
...
 Now you can access the served content at `http://mysite.local:4568/`
 
+## Subscribing to S3 Event 
+
+You can subscribe to Put, Copy and Delete object events in the bucket.
+Please have a look at [Aws page](http://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html) for details of event object. 
+Apply filter function to subscribe to specific events.
+
+```
+var S3rver = require('s3rver');
```
leontastic: Please update this to `const`, we'd like to promote ES6 syntax were possible (even though a lot of our older code uses `var`)

lib/models/s3-event.js
```diff
@@ -0,0 +1,83 @@
...
+    const authorization = reqHeaders.authorization.split(':');
+    principal = authorization[0].split(' ');
+  } catch (err) {
+    principal = ['AWS', randomString(21).toUpperCase()];
+  }
+  var eventName = '';
```
leontastic: Use `let` instead of `var`

lib/models/s3-event.js
```diff
@@ -0,0 +1,83 @@
...
+  var s3Object = {
```
leontastic: Use `const` instead of `var`

lib/models/s3-event.js
```diff
@@ -0,0 +1,83 @@
...
+          "object": s3Object
+        }
+      }
+    ]
+  }
+  return event;
+};
+
+const randomString = function (length) {
+  var randString = "";
```
leontastic: Use `let` instead of `var`

lib/models/s3-event.js
anantab: It's been month and I forgot why I did it that way before.  I have removed that block anyways as I don't see any principalId information in request headers.

lib/models/s3-event.js
anantab: Hi, the real s3 event object has "AWS:" prepended in principalId.

bin/s3rver.js
anantab: I have removed the code from this file. With this feature, you can only subscribe to events when you run s3rver programatically as you need to subscribe to s3rver.s3Event observable. Not sure how you would do that if you run s3rver from command line. Any suggestions? Pass a function as parameter?

lib/models/s3-event.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+const crypto = require('crypto');
+
+const S3Event = function (eventData, reqParams) {
+  const {reqHeaders,sourceIp} = reqParams;
+  let eventName = '';
+  const s3Object = {
+    "key": eventData.S3Item.key,
+    "sequencer": ('00' + (Math.floor(Date.now())).toString(16)).toUpperCase()
```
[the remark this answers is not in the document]
leontastic: According to the [notifications docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html): 

> The sequencers can be of different lengths. So to compare these values, you first right pad the shorter value with zeros and then do lexicographical comparison.

Since there's no guarantee of the sequencer length, you could just use `Date.now().toString(16).toUpperCase()` without prepending `00` and it would match the specified behaviour.

lib/models/s3-event.js
```diff
@@ -0,0 +1,79 @@
...
+  const event = {
+    "Records": [
+      {
+        "eventVersion": "2.0",
+        "eventSource": "aws:s3",
+        "awsRegion": reqHeaders.host,
+        "eventTime": eventData.S3Item.creationDate || new Date().toISOString(),
+        "eventName": eventName,
+        "userIdentity": {
+          "principalId": 'AWS:' + randomString(21).toUpperCase(),
```
leontastic: The `userIdentity.principalId` field shown in the [example](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html) isn't prepended with `AWS:`. Please remove the prefix.

lib/models/s3-event.js
```diff
@@ -0,0 +1,79 @@
...
+        "requestParameters": {
+          "sourceIPAddress": sourceIp
+        },
+        "responseElements": {
+          "x-amz-request-id": randomString(16).toUpperCase(),
+          "x-amz-id-2": crypto.createHash('sha256').update(reqHeaders.host).digest('base64')
+        },
+        "s3": {
+          "s3SchemaVersion": "1.0",
+          "configurationId": randomString(8) + '-' + randomString(4) + '-' + randomString(4) + '-' + randomString(4) + '-' + randomString(12),
```
leontastic: This is specified as the ID of the specified bucket notification configuration ([example](https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html)), since we don't support notification configs right now can we define this as `testConfigId` instead of the arbitrary ID format you've used here?

lib/models/s3-event.js
```diff
@@ -0,0 +1,79 @@
...
+          },
+          "object": s3Object
+        }
+      }
+    ]
+  }
+  return event;
+};
+
+const randomString = function (length) {
```
leontastic: Apologies, I should have pointed this out in my last review. Instead of this function, can we use `crypto` for simplicity? Here's an example: https://stackoverflow.com/a/40191779/4743864

lib/models/s3-event.js
```diff
@@ -0,0 +1,79 @@
...
+      eventName = 'ObjectRemoved:Delete';
+      break;
+  }
+
+  const event = {
+    "Records": [
+      {
+        "eventVersion": "2.0",
+        "eventSource": "aws:s3",
+        "awsRegion": reqHeaders.host,
```
leontastic: Is this correct? Would rather specify a constant `us-east-1` than use an incorrect value.

bin/s3rver.js
leontastic: You could add another command line argument above in this file, and create the subscription only if the argument is set. Would need to add a little more documentation to make the feature accessible to users.

anantab: that would not help because users would not get event notification in their project. May be we could use remote observable (https://github.com/jbaudanza/rxremote) to subscribe to server over websocket. Something may be for next release.

PR #126, opened 2018-02-11
Add eslint and prettier

n1ru4l:


n1ru4l:
@leontastic @specialkk What do  you think about precommit hooks? I am a fan of having one for prettier only.

specialkk:
I don't see any real problem with that since Travis will be running the linter anyway. However what's your reasoning for not including the check up-front in the precommit hook?

n1ru4l:
@specialkk To see clean diffs for in progress pull requests, and I do not want to be forced to fix eslint errors for an "wip" commit to succeed.

PR #131, opened 2018-02-14
Fix copyObject operations with "COPY" MetadataDirective

specialkk:
I've actually encountered this and worked around it thinking it was just some annoying behavior for S3, but it's actually just been implemented wrong here. Ideally this would have been caught a long time ago since it's basically the result of a syntax error, but `var` is just the best at hiding those -_-.

I'm still working on rebasing the fs store refactor I've had for a while now that has a much better implementation of this with proper error handling, but I'll add this now just since there's some value in giving new test cases their own PR.

### DIFF
--- a/lib/file-store.js
+++ b/lib/file-store.js
@@ -343,18 +343,18 @@ const FileStore = function(rootDirectory) {
   };
 
   const copyObject = function(options, done) {
-    var req = options.request,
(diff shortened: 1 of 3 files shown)


leontastic:
@specialkk Is this good to go?

specialkk:
Yes it is!

PR #134, opened 2018-02-16
(fix) cors.js has a typo for `.toLowerCase()`

misterdai:


leontastic:
@misterdai Good catch!

PR #138, opened 2018-02-16
Multipart form data support

recunius:
I added support for multipart/form-data POST to S3 (see https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html for an example).

This adds a dependency on the multer library.

I also made an update to the lib/cors.js file. I was running into problems because the express request.get(...) call was returning a string for the header value but s3rver is then treating it as an array. I'm not confident enough to say req.get(...) will always return a string, but according to the express docs it looks that way: http://expressjs.com/en/api.html

Limitations:
 * not all X-Amx-* form fields are applied
 * does not check S3 POST policy


specialkk:
Hey, thanks for your contribution! I've created and merged separate PR at #140 regarding the CORS issue, it was indeed a minor oversight on my part. Please rebase onto master when you can.

I'd definitely like some tests for this. Everything else looks pretty good, I might request some minor changes but I don't see any major problems right now.

leontastic:
@recunius This change looks pretty good. Thank you for your contribution!

Can you add some test cases of multipart form uploads to prevent future regressions, and update the docs to include this feature?

recunius:
@specialkk @leontastic Hi, yes, I started working on the test cases (not done yet). I'll update the README as well. It'll probably take me a bit of time to get around to it fully.

I do warn that even once complete the multipart/form-data POST will not match all of the features present on the true AWS S3 platform. I don't personally have the resources to work towards that goal.

PR #139, opened 2018-02-17
Use modern ES features where applicable

specialkk:
- Use classes where applicable
- Use arrow functions for callbacks
- Use object initializer shorthands
- Swap lodash functions with native implementations where possible
- Alphabetically order imported modules

I avoided making everything use Promises since I'd rather jump straight to async/await style when we eventually decide to do that.

This shouldn't change any existing behavior at all, it should all be purely syntax updates.

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -1,28 +1,39 @@
 "use strict";
 
-const FileStore = require("./file-store");
-const templateBuilder = require("./xml-template-builder");
-const concat = require("concat-stream");
-const xml2js = require("xml2js");
 const async = require("async");
-const path = require("path");
-const ReadableStream = require("stream").Readable;
+const concat = require("concat-stream");
 const crypto = require("crypto");
+const path = require("path");
+const { Readable: ReadableStream } = require("stream");
 const url = require("url");
+const xml2js = require("xml2js");
+
+const FileStore = require("./file-store");
 const S3Event = require("./models/s3-event");
+const templateBuilder = require("./xml-template-builder");
 
 module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
   const fileStore = new FileStore(rootDirectory);
 
-  const buildXmlResponse = function(res, status, template) {
+  function buildXmlResponse(res, status, template) {
     res.header("Content-Type", "application/xml");
     res.status(status);
     return res.send(template);
-  };
+  }
 
-  const buildResponse = function(req, res, status, object, data) {
-    res.header("Etag", '"' + object.md5 + '"');
-    res.header("Last-Modified", new Date(object.modifiedDate).toUTCString());
+  function buildResponse(req, res, status, object, data) {
+    res.header("Accept-Ranges", "bytes");
+    if (data.range) {
+      const end = Math.min(data.range.end || Infinity, object.size - 1);
+      res.header(
+        "Content-Range",
+        "bytes " + data.range.start + "-" + end + "/" + object.size
+      );
+      res.header("Content-Length", end - data.range.start + 1);
+    } else {
+      res.header("Content-Length", object.size);
+    }
+
     res.setHeader("Content-Type", object.contentType);
 
     if (object.contentEncoding)
@@ -31,40 +42,33 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
     if (object.contentDisposition)
       res.header("Content-Disposition", object.contentDisposition);
 
-    if (data.range) {
-      var end = Math.min(data.range.end || Infinity, object.size - 1);
-      res.header(
-        "Content-Range",
-        "bytes " + data.range.start + "-" + end + "/" + object.size
-      );
-      res.header("Accept-Ranges", "bytes");
-      res.header("Content-Length", end - data.range.start + 1);
-    } else {
-      res.header("Content-Length", object.size);
-    }
+    res.header("Etag", JSON.stringify(object.md5));
+    res.header("Last-Modified", new Date(object.modifiedDate).toUTCString());
 
     if (object.customMetaData.length > 0) {
-      object.customMetaData.forEach(function(metaData) {
-        res.header(metaData.key, metaData.value);
-      });
+      for (const { key, value } of object.customMetaData) {
+        res.header(key, value);
+      }
     }
     res.status(status);
-    if (req.method === "HEAD") {
-      return res.end();
-    }
-    return data.pipe(res);
-  };
 
-  const triggerS3Event = function(req, res, eventData) {
+    if (req.method === "HEAD") {
+      res.end();
+    } else {
+      data.pipe(res);
+    }
+  }
+
+  function triggerS3Event(req, res, eventData) {
     res.app.s3Event.next(
       new S3Event(eventData, {
         reqHeaders: req.headers,
         sourceIp: req.connection.remoteAddress
       })
     );
-  };
+  }
 
-  const errorResponse = function(req, res, keyName) {
+  function errorResponse(req, res, keyName) {
     logger.error(
       'Object "%s" in bucket "%s" does not exist',
       keyName,
@@ -73,11 +77,7 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
 
     if (indexDocument) {
       if (errorDocument) {
-        fileStore.getObject(req.bucket, errorDocument, function(
-          err,
-          object,
-          data
-        ) {
+        fileStore.getObject(req.bucket, errorDocument, (err, object, data) => {
           if (err) {
             // eslint-disable-next-line no-console
             console.error("Custom Error Document not found: " + errorDocument);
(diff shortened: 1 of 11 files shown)


lib/file-store.js
```diff
@@ -333,172 +323,120 @@ const FileStore = function(rootDirectory) {
...
     });
-    const partStreams = _.map(partPaths, function(partPath) {
-      return fs.createReadStream(path.join(partPath, CONTENT_FILE));
-    });
-    const combinedPartsStream = concatStreams(new PassThrough(), partStreams);
+    const partStreams = partPaths.map(partPath =>
+      fs.createReadStream(path.join(partPath, CONTENT_FILE))
+    );
+    const combinedPartsStream = this.concatStreams(
+      new PassThroughStream(),
```
leontastic: @specialkk I don't have a strong opinion about this but why assign your own name to `PassThrough`? The original name is clear enough and common to Node.js, reading this made me think that you rolled out your own version.

lib/index.js
```diff
@@ -1,69 +1,70 @@
...
-      function(bucket, callback) {
-        fs.remove(path.join(directory, bucket), callback);
-      },
-      callback
-    );
-  });
-};
+class S3rver {
+  constructor(options) {
+    this.options = Object.assign({}, S3rver.defaultOptions, options);
```
leontastic: @specialkk This is not the same as the previous code. An `undefined` property in `options` would override `defaultOptions` here, whereas the default would be used in the previous code.

I'm assuming you want to have fewer dependencies, but there are benefits to using `lodash` that outweigh the cost of depending on it in 99% of the cases (even `forEach` – lodash is faster than native). Destructure the methods if you want to get rid of the ugly `_.` syntax but I'd encourage you to use `lodash` wherever it makes sense, and not to remove lodash method calls with prejudice.

lib/models/s3-event.js
```diff
@@ -1,74 +1,75 @@
...
-  const s3Object = {
-    "key": eventData.S3Item.key,
-    "sequencer": Date.now().toString(16).toUpperCase()
-  };
-  switch (eventData.eventType) {
-    case 'Copy':
-      eventName = 'ObjectCreated:Copy';
-      s3Object.size = eventData.S3Item.size;
-      break;
+function randomString(length) {
```
leontastic: Arrow function?

lib/utils.js
```diff
@@ -23,25 +24,21 @@ exports.walk = function(dir) {
...
-function removeEmptyDirectories(fs, directory, callback) {
-  fs.stat(directory, function(err, stat) {
-    if (err) return callback(err);
-    if (!stat.isDirectory()) return callback();
-    fs.readdir(directory, function(err, list) {
-      if (err) return callback(err);
-      if (list.length === 0) {
-        return fs.rmdir(directory, callback);
-      }
+exports.removeEmptyDirectories = function(fs, directory, done) {
```
leontastic: Arrow function?

lib/xml-template-builder.js
```diff
@@ -1,236 +1,245 @@
...
-        return {
-          CommonPrefixes: { Prefix: prefix }
-        };
-      })
-    );
-    content.unshift({
+const DISPLAY_NAME = "S3rver";
+
+function buildQueryContentXML(data, options) {
+  const content = [
```
leontastic: Could probably express this without needing to call `push`:

```javascript
const content = [
  { /* ... */ },
  ...data.objects.map(/* ... */),
  ...data.commonPrefixes.map(/* ... */)
]
```

test/test.js
```diff
@@ -1164,12 +1067,10 @@ describe("S3rver Tests", function() {
     }
     async.eachSeries(
       testObjects,
-      function(testObject, callback) {
-        s3Client.putObject(testObject, function(err, object) {
+      (testObject, callback) => {
+        s3Client.putObject(testObject, (err, object) => {
+          if (err) return done(err);
```
leontastic: Probably still want to call `callback` instead of `done` here so that the `eachSeries` doesn't continue after the test fails.

test/test.js
```diff
@@ -1710,11 +1596,9 @@ describe("Data directory cleanup", function() {
         sslEnabled: false,
         s3ForcePathStyle: true
       });
-      s3Client.createBucket({ Bucket: bucket }, function(err) {
-        if (err) return done(err);
-        generateTestObjects(s3Client, bucket, 10, function(err) {
-          if (err) return done(err);
-          s3rver.close(function(err) {
+      s3Client.createBucket({ Bucket: bucket }, err => {
+        generateTestObjects(s3Client, bucket, 10, () => {
```
leontastic: Why was `if (err) return done(err)` removed for these two calls? Two instances of this change below as well.

lib/controllers.js
```diff
@@ -1,70 +1,74 @@
...
-    res.setHeader("Content-Type", object.contentType);
-
-    if (object.contentEncoding)
-      res.header("Content-Encoding", object.contentEncoding);
-
-    if (object.contentDisposition)
-      res.header("Content-Disposition", object.contentDisposition);
+  }
 
+  function buildResponse(req, res, status, object, data) {
```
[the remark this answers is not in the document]
specialkk: That sounds fine. I just like how function declarations stick out and are bit easier to spot when skimming through code but I really can go either way.

lib/index.js
specialkk: You're right, I'll revert the `Object.assign`. Nothing against using lodash. I think the only instances I ended up replacing it was when `_.map` was used.

lib/utils.js
specialkk: This is the only place where I do have opinion on this. It's a top-level function, so the `this` context will never be relevant. With ESM modules, I also just prefer how
```javascript
export function removeEmptyDirectories(fs, directory, done) { ... }
```
looks compared to
```javascript
export const removeEmptyDirectories = (fs, directory, done) => { ... }
```
The only place I actually prefer arrow functions is when they are passed directly into callbacks, since those are the only places I've found that `this` is actually relevant (with a few exceptions).

If this seems too inconsistent I can update it everywhere, there doesn't need to be much debate.

test/test.js
specialkk: Wow, good eye!

lib/xml-template-builder.js
specialkk: Since you bring that up, this whole function can be inlined since it's only used in `buildBucketQuery`.

lib/file-store.js
specialkk: I've just developed the habit of renaming them to be suffixed with `Stream` ever since destructuring imports became common. Before you'd at least be able to see
```javascript
var s = new stream.PassThrough();
```
but now it looks like
```javascript
const s = new PassThrough();
```
It's kind of a flimsy reason and I can see how it can look weird so I'll revert it.

lib/file-store.js
```diff
@@ -1,133 +1,183 @@
...
+  }
 
-  const deleteBucket = function(bucket, done) {
-    const bucketPath = getBucketPath(bucket.name);
-    fs.rmdir(bucketPath, function(err) {
-      if (err) {
-        return done(err);
-      }
-      return done();
+  concatStreams(passThrough, streams) {
```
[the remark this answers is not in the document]
specialkk: In this particular case, I've actually already moved `concatStreams` to utils.js in the filesystem store refactor I'll be submitting soon! I was also able to simplify things enough to the point where there's only a couple functions left in the "helpers" section.

I'm also actually envisioning documenting the public interface needed for other potential storage backends. Ideally other functions that aren't part of this interface could be marked private, but I'm okay with having a document that pretty explicitly states "these are the methods this class is required to implement".

lib/file-store.js
```diff
@@ -333,172 +323,120 @@ const FileStore = function(rootDirectory) {
...
-            return done("Error updating metadata");
-          }
-          return done(null, new S3Object(metaData));
+        (err, metaData) => {
+          if (err) return done("Error updating metadata");
+          done(null, new S3Object(metaData));
         }
       );
-    } else {
+    } else if (srcKeyPath !== destKeyPath) {
```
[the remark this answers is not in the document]
specialkk: This is a relic of me messing around for the fix I made in #131 that I accidentally left in here. `controllers.js` should actually be checking to make sure that the source and destination aren't the same since there's a specific error response S3 will send if you forget to specify the REPLACE MetadataDirective. I'll properly address this in an upcoming PR, but I'll revert this for now.

lib/controllers.js
```diff
@@ -114,21 +114,18 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+  }
 
-  const deleteObjects = function(req, res) {
-    xml2js.parseString(req.body, function(err, parsedBody) {
-      const keys = parsedBody.Delete.Object.map(function(o) {
-        return o.Key[0];
-      });
+  function deleteObjects(req, res) {
+    xml2js.parseString(req.body, (err, parsedBody) => {
+      const keys = parsedBody.Delete.Object.map(o => o.Key[0]);
```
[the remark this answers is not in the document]
specialkk: For situations like this I usually only bother when there's multiple properties I want to access. Right here I guess I should either use destructuring or give `o` a more meaningful name.

specialkk:
@leontastic I've addressed everything I wanted to so far, I just wanted some extra input on what to do about the arrow functions.

PR #146, opened 2018-02-22
Add passing test for #62

### LINKED ISSUE #62: Copying from bucket to bucket fails when the path contains a colon
There appears to be a bug in the fake S3 server implementation whereby the applciation is not decoding the HTTP message headers (only the URL) when handling the S3 copy command.
The following script works against the real AWS S3 but fails to copy from bucket to bucket in your fake implementation with this error:

```
Copying file from s3://bob-test-bucket-1/test-2017-03-09T10:54:24.txt to s3://bob-test-bucket-2/test-2017-03-09T10:54:24.txt
copy failed: s3://bob-test-bucket-1/test-2017-03-09T10:54:24.txt to s3://bob-test-bucket-2/test-2017-03-09T10:54:24.txt An error occurred (NoSuchKey) when calling the CopyObject operation: The specified key does not exist
```

And here’s the script to test it:

```bash
#!/bin/bash

export OVERRIDE_URL="--endpoint-url http://localhost:4569"

echo "THIS IS A TEST" > test-$$.txt

echo "Listing buckets"
aws s3 $OVERRIDE_URL ls
echo

echo "Creating bucket bob-test-bucket-1"
aws s3 $OVERRIDE_URL mb s3://bob-test-bucket-1
echo

echo "Creating bucket bob-test-bucket-2"
aws s3 $OVERRIDE_URL mb s3://bob-test-bucket-2
echo

export SUFFIX=$(date -u "+%Y-%m-%dT%H:%M:%S")

echo "Uploading test data file to s3://bob-test-bucket-1/test-$SUFFIX.txt"
aws s3 $OVERRIDE_URL cp test-$$.txt s3://bob-test-bucket-1/test-$SUFFIX.txt
echo

echo "Copying file from s3://bob-test-bucket-1/test-$SUFFIX.txt to s3://bob-test-bucket-2/test-$SUFFIX.txt"
aws s3 $OVERRIDE_URL cp s3://bob-test-bucket-1/test-$SUFFIX.txt s3://bob-test-bucket-2/test-$SUFFIX.txt
echo

rm test-$$.txt

echo "Done!"
```

n1ru4l:
Closes #62

Either I am not understanding the issue or it is working as it is supposed to 😅

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -561,6 +561,45 @@ describe("S3rver Tests", function() {
     });
   });
 
+  it("should copy an image object whose key includes a colon into another bucket", function(done) {
+    const srcKey = "test-2017-03-09T10:54:24.jpg";
+    const destKey = "image/test-2017-03-09T10:54:24.jpg";
+    const file = path.join(__dirname, "resources/image.jpg");
+
+    fs.readFile(file, function(err, data) {
+      if (err) {
+        return done(err);
+      }
+      const params = {
+        Bucket: buckets[0],
+        Key: srcKey,
+        Body: new Buffer(data),
+        ContentType: "image/jpeg",
+        ContentLength: data.length
+      };
+      s3Client.putObject(params, function() {
+        const params = {
+          Bucket: buckets[3],
+          Key: destKey,
+          CopySource: "/" + buckets[0] + "/" + srcKey
+        };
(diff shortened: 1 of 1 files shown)


n1ru4l:
Yeah, just wanted to prove the world it is actually working 😂 I am also fine with just closing the issue and this PR.

Issue #154, opened 2018-02-27
Error when object has no content-type


n1ru4l:
If you request a resource that has no content-type stored in its `.dummys3_metadata` the server throws an unhandled error.

AWS fallsback to the `binary/octet-stream` content-type when there is none specified (https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html).

If you upload a file to s3 via the web interface it guesses the correct file type.

This issue occurs if you do a PUT operation using node-fetch or curl without specifying the `Content-Type` header.

PR #156, opened 2018-02-28
refactor(FilesystemStore): Cleanup and fixes

### LINKED ISSUE #38: Header Options
Heyho, is it planned to add more options (e.g. header options like 'cache-control', as for example defined here: http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#copyObject-property) to the aws-functions as well as the according controllers?

I am manually setting the 'Cache-control' option on a couple of objects and would like to be able to test for the correct response. 

Thanks for your time and good work!

specialkk:
The primary goal here is to decouple the controller functions from the logic needed to prepare object data for reading/writing. This should make it easy to implement alternate storage backends, e.g. in-memory storage or MongoDB GridFS.

Features that got fixed/implemented along the way:
* All metadata types should now be supported (#38)
* Make copyObject behave more like real S3
  * All metadata (including content type) correctly gets replaced now
  * MD5 hash is stored separately to avoid being replaced by metadata updates
  * Invalid requests are correctly handled
* Support the isTruncated flag for listObjects
* Better handling of listObjects `marker` parameter
* All tests on Windows seem to work now

This should close #38 and #154, still needs tests though.


specialkk:
Not sure what was going wrong I my end that I wasn't able to push a working version, but it looks I finally got it to work.

Anyway, one thing I had to work around was preserving the existing format for storing object metadata. Ideally I'd like to just dump the headers as-is without having to translate them to/from the property names they have in `.dummys3_metadata` JSON. Also, I'd like to maybe rename the files to `.s3rver_content` and `.s3rver_metadata.json`.

I'm assuming we should wait until another major release to break compatibility with existing data folders? Would an upgrade script through the CLI be acceptable for a minor release?

specialkk:
@leontastic any comments? I know you usually find manage to find something that needs adjusted.

The only other thing I currently plan to do with this PR at the moment is add a test/fix that closes #154. I think #38 is safe to close unless we want another PR that adds tests for every kind of header.

test/test.js
```diff
@@ -455,7 +455,7 @@ describe("S3rver Tests", function() {
             (err, object) => {
               if (err) done(err);
               object.Metadata.should.have.property("somekey", "value");
-              object.ContentType.should.equal("image/jpeg");
+              object.ContentType.should.equal("application/octet-stream");
```
n1ru4l: Did you check this behaviour on s3?

I am not 100% sure if s3 tries to guess the header or not before adding `application/octet-stream`.

test/test.js
```diff
@@ -496,7 +496,7 @@ describe("S3rver Tests", function() {
             (err, object) => {
               if (err) return done(err);
               object.Metadata.somekey.should.equal("value");
-              object.ContentType.should.equal("image/jpeg");
+              object.ContentType.should.equal("application/octet-stream");
```
n1ru4l: Same as above

lib/controllers.js
```diff
@@ -72,20 +65,26 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+          errorDocument,
+          (err, object, data) => {
+            if (err) {
+              // eslint-disable-next-line no-console
+              console.error(
+                "Custom Error Document not found: " + errorDocument
+              );
+              return notFoundResponse(req, res);
+            } else {
+              return buildResponse(req, res, 404, object, data);
```
leontastic: This call to `buildResponse` is still using the old function signature (5 params instead of 4 in your changes).

lib/controllers.js
```diff
@@ -389,71 +415,58 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
     putObject(req, res) {
-      if (req.query.uploadId) {
-        return putObjectMultipart(req, res);
-      }
-
       if (req.headers["x-amz-copy-source"]) {
-        handleCopyObject(req.params.key, req, res);
+        copyObject(req, res);
+      } else if (req.query.uploadId) {
+        putObjectMultipart(req, res);
```
leontastic: 👍 Good job factoring out the `if (req.headers["x-amz-copy-source"])` branch from `putObjectMultipart`

lib/controllers.js
```diff
@@ -464,32 +477,46 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+            deleteObjects(req, res);
+          })
+        );
+      } else {
+        notFoundResponse(req, res);
+      }
+    },
+
     postObject(req, res) {
+      let template;
```
leontastic: I know we follow this pattern elsewhere in the code, but there's no need for `template` to be mutable since we pass it into `buildXmlResponse` and return everywhere we set `template`. Instead, use `const template` right before calling `buildXmlResponse` to reduce the cognitive load of reading this code.

lib/stores/filesystem.js
```diff
@@ -0,0 +1,382 @@
...
+            "content-language": storedMetadata.contentLanguage,
+            "content-type": storedMetadata.contentType || "application/octet-stream",
+            "expires": storedMetadata.expires,
+            "website-redirect-location": storedMetadata.websiteRedirectLocation,
+
+            "last-modified": stats.mtime.toUTCString(),
+            "etag": JSON.stringify(md5),
+            "content-length": stats.size
+          },
+          isUndefined
```
leontastic: 👍 I love this pattern `omitBy(object, isUndefined)`

lib/stores/filesystem.js
```diff
@@ -0,0 +1,382 @@
...
+
+    if (options.prefix) {
+      keys = keys.filter(key => key.startsWith(options.prefix));
+    }
+
+    if (options.delimiter) {
+      const prefix = options.prefix || "";
+      keys = keys.filter(key => {
+        const idx = key.slice(prefix.length).indexOf(options.delimiter);
+        if (!~idx) return true;
```
leontastic: I had to Google this `!~idx` to understand what you were trying to do. Pretty clever, but think there are more readable ways to express this. [`String.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) or [`lodash.includes`](https://lodash.com/docs/4.17.5#includes) would make this logic look closer to English.

specialkk: I guess it might be a bit obscure but I've found it to be pretty readable once you're used to it. (If you want something really funky look up using `~~( )` as an efficient floor function!)

I'll just convert it to `idx !== -1` to help readability. I don't really want to do `includes` since I need to know the index if it does exist.

specialkk:
@leontastic let me know if this looks good. @n1ru4l once this gets merged I'll see about making a PR to close #154.

AlixBa:
hi guys, do you have an ETA for this?

PR #163, opened 2018-03-05
Support running on port 0

### LINKED ISSUE #162: Support running on any available port (port 0)
I want to instantiate many `s3rver` instances within my [AVA](https://github.com/avajs/ava) tests, where they will run concurrently, with some in their own process.

The best I have been able to do so far is increment a counter for the port number so the different instances do not conflict with each other. But this is brittle because using a large contiguous range of port numbers increases the chances of conflict with an existing service on a developer's machine.

Ideally, I should be able to just run `s3rver` on [port 0](https://www.lifewire.com/port-0-in-tcp-and-udp-818145) to dynamically allocate an available port.

```js
const server = new S3rver({
    directory : await mkdirtemp(),
    hostname  : 'localhost',
    port      : 0,
    silent    : true
});
server.run((err, hostname, port) => {
    const endpoint = `http://${hostname}:${port}`;
    // ... create S3 client that connects to endpoint ...
});
```

I figured it would work, as most servers support that, so I was surprised to see this error:

```js
NetworkingError (Error) {
    address: '127.0.0.1',
    code: 'NetworkingError',
    errno: 'ECONNREFUSED',
    hostname: 'localhost',
    message: 'connect ECONNREFUSED 127.0.0.1:80',
    port: 80,
    region: 'us-east-1',
    retryable: true,
    syscall: 'connect',
    time: Date 2018-03-05 07:54:29 704ms UTC {},
}
```

After investigation, it seems my AWS client tried to connect on port 80 because `s3rver` just returns the port as-is instead of the allocated port, and port 0 is invalid when making a request. The expected behavior is that it would use the return value of [server.address()](https://nodejs.org/api/net.html#net_server_address).

sholladay:
When the user passes `port : 0`, the operating system will dynamically allocate an available port for us, which we then return to the user so that they know how to connect to us.

Fixes #162

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -43,7 +43,7 @@ class S3rver {
         done(
           err,
           this.options.hostname,
-          this.options.port,
+          server.address().port,
           this.options.directory
         );
       })
--- a/test/test.js
+++ b/test/test.js
@@ -1605,6 +1605,17 @@ describe("S3rver Class Tests", function() {
     s3rver.options.cert.should.be.an.instanceOf(Buffer);
     s3rver.options.should.have.property("removeBucketsOnClose", true);
   });
+  it("should support running on port 0", function(done) {
+    const s3rver = new S3rver({
+      port: 0,
+      hostname: "localhost",
+      silent: true
+    }).run((err, hostname, port) => {
+      if (err) return done(err);
+      should(port).be.above(0);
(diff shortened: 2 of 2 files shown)


leontastic:
@sholladay Thank you for your contribution!

PR #166, opened 2018-03-07
Fix issue where deleting the last top level object also deletes the bucket

electrikdevelopment:
Updated removeEmptyDirectories so that the top level directory passed is not deleted itself.


leontastic:
@electrikdevelopment Thank you for your contribution!

PR #168, opened 2018-03-07
Marker should be a prefix, not a full match

AlixBa:
Our s3 production works this way so I guess this code hasn't been updated or has always been false.
I would love a patch release for this <3

### DIFF
--- a/lib/file-store.js
+++ b/lib/file-store.js
@@ -232,7 +232,7 @@ class FileStore {
       let startAt = 0;
       let found = false;
       filteredKeys.forEach((key, index) => {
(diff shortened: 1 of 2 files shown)


specialkk:
Hey, I think this is already fixed in a PR that should get merged pretty soon (#156). You can check the [new implementation here](https://github.com/jamhall/s3rver/blob/fs-refactor/lib/stores/filesystem.js#L201).

Once that merges, we'd definitely still appreciate improvements to the test suite. I'll keep you updated.

AlixBa:
Seems to be what I'm looking for, indeed. I did not check this PR as it was labelled "clean" :D

PR #170, opened 2018-03-08
Fix script globs

specialkk:
I've noticed that Travis hasn't been picking up formatting adjustments that I keep picking up in my Windows env, this fix should hopefully have it catch those in future PRs.


jamhall:
@specialkk thank you for your excellent work and contributions!

Issue #155, opened 2018-02-27
Don't extend object prototype in tests


n1ru4l:
Can we please use expect over should and extending the object prototypes?

leontastic:
I support this proposal, I prefer explicit assertions. Seeing `expect` at the beginning of the line lets me immediately know I'm looking at an assertion.

PR #175, opened 2018-03-09
refactor(test): Use chai-expect instead of should

specialkk:
I accidentally pushed to master initially 😬, sorry about that.

Anyway this is for #155 since extending the Object prototype isn't that great.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "devDependencies": {
     "aws-sdk": "2.207.0",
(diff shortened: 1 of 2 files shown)


n1ru4l:
We can lock master to prevent such things 😂

Could you please resolve the conflicts?

Otherwise I love what you are doing 😊

specialkk:
Thanks! 😊

I'll go ahead and merge this before the renovate bot causes more conflicts.

PR #177, opened 2018-03-09
Use co-mocha for async tests

specialkk:
This improves error handling within tests so that they shouldn't time out when an assertion fails from within a callback. Server instances created within tests are also closed in finally blocks to avoid making Mocha hanging at the end.

This also greatly improves the performance of bulk generation of objects by allowing 100 concurrent uploads at a time. The indefinite timeout on these long-running tests has been removed.

edit: wooo new record of only 3:28 for total test runtime on Travis!


specialkk:
Cool, I'll merge now since it looks like we'll need to be adding some more tests pretty soon.

PR #189, opened 2018-03-15
Publish a 2.2.2 patch version

AlixBa:


n1ru4l:
cc @leontastic @jamhall

PR #289, opened 2018-08-02
Send date when triggering S3 Event after Post/Put

nschirmer:
S3 Events that are triggered do not contain a proper eventTime value, because of how the code is set up. They instead have "Invalid Date" objects.

There's a few ways this could probably be fixed, one of which is to actually send a date value when triggering the S3 event from the controller after Put/Post object.

Also, "date" wasn't even a proper metadata key within s3-object, so I've added that.

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -500,7 +500,8 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
             eventType: "Put",
             S3Item: new S3Object(object.bucket, object.key, null, {
               "content-length": size,
-              etag: JSON.stringify(md5)
+              etag: JSON.stringify(md5),
+              date: new Date().toISOString()
             })
           });
           res.header("ETag", JSON.stringify(md5));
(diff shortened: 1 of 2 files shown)


nschirmer:
@jamhall Let me know if this makes sense -- I'm building a Serverless lambda for S3 events and my code is erroring when testing locally in Serverless Offline unless I add some extra code for expecting a bad `event.Records[0].eventTime` value.

specialkk:
Hey, thanks for pointing this out. I'm fairly certain that getter for `creationDate` in the `S3Object` class was left in by mistake. AFAIK there aren't any part of S3 that needs a key's original creation date, and I'm also not sure why the original PR for S3 events uses the creation date for the `eventTime`.

https://github.com/jamhall/s3rver/blob/f61fb9cae55f471a982678ed01776f10a2f82fc1/lib/models/s3-event.js#L45-L50

nschirmer:
@specialkk Good point! This PR could be simplified then and, instead of the changes I made, simply make the `eventTime` always use a new date.

PR #325, opened 2018-10-08
CLI: Exit with non-zero status with misconfigured data directory

kherock:
Self-explanatory title. For me exiting with a status of zero is problematic when configuring S3rver for use as a background task in VSCode.


leontastic:
@kherock Thank you! Released in v2.2.7:
https://github.com/jamhall/s3rver/releases/tag/v2.2.7

PR #333, opened 2018-10-30
Generic post method to actually allow uploading files

codyseibert:
This PR adds support for uploading a file directly to the bucket via a multipart form post.

### Overview
The createPresignedPost method is an AWS s3 method which creates a presigned url and fields which can be used for posting an s3 object directly to a private bucket

An example of calling this method:
```
s3.createPresignedPost({
  Bucket: 'my_bucket',
  Conditions: [
    ['starts-with', '$key', ''],
  ],
}, (err, data) => {
  if (err) return reject(err);
  resolve(data);
});
```

which returns something like this:
```
{
  "url": "http://localhost:9000/my_bucket",
  "fields": {
    "bucket": "my_bucket",
    "X-Amz-Algorithm": "AWS4-HMAC-SHA256",
    "X-Amz-Credential": "XXX/us-east-1/s3/aws4_request",
    "X-Amz-Date": "20181030T043058Z",
    "X-Amz-Security-Token": "XXX",
    "Policy": "XXX",
    "X-Amz-Signature": "XXX"
  }
}
```

Normally, you can just use the URL returned in the response in a multipart form and upload the file to the bucket.  Currently, s3rver has no support for this upload method.  This PR adds that support.

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -519,7 +519,7 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
           })
         );
       } else if (
-        req.headers["content-type"].indexOf("multipart/form-data") !== -1
+        (req.headers["content-type"] || req.headers["Content-Type"]).indexOf("multipart/form-data") !== -1
       ) {
         const fields = {
           bucket: req.params.bucket
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -519,7 +519,9 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
           })
         );
       } else if (
-        (req.headers["content-type"] || req.headers["Content-Type"]).indexOf("multipart/form-data") !== -1
+        (req.headers["content-type"] || req.headers["Content-Type"]).indexOf(
+          "multipart/form-data"
+        ) !== -1
       ) {
         const fields = {
           bucket: req.params.bucket
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -7,6 +7,7 @@ const path = require("path");
 const { Readable } = require("stream");
 const url = require("url");
 const xml2js = require("xml2js");
+const multiparty = require("multiparty");
 
 const S3Event = require("./models/s3-event");
 const S3Object = require("./models/s3-object");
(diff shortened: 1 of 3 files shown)


leontastic:
@codyseibert Thank you for taking the time to contribute a PR! Can you add some test cases for this functionality? I'm happy to review this PR, but we need tests proving your change provides the expected results you described.

codyseibert:
@leontastic is this PR good now sir?

lib/controllers.js
```diff
@@ -517,6 +518,88 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+                part
+                  .on("data", buf => (string += buf.toString()))
+                  .on("end", () => {
+                    fields[part.name] = string;
+                    resolve();
+                  });
+              });
+            });
+          } else {
+            // 'file' must be the last field in the post (this is how aws requires it)
```
leontastic: Can you elaborate on this comment? I don't see any reference to a `file` field in the code

lib/controllers.js
```diff
@@ -517,6 +518,88 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+              }
+              store.putObject(
+                {
+                  bucket: fields.bucket,
+                  key: fields.key,
+                  metadata: {
+                    "content-type":
+                      fields["Content-Type"] || fields["content-type"]
+                  },
+                  content: part
```
leontastic: Is it possible for a file upload to be spread across multiple parts? If so then this will cause subsequent parts to overwrite previous parts stored at the same bucket and key, in parallel because `putObject` is async. You would need to concatenate the multiple `ReadableStream` parts together before calling `putObject` just once.

I couldn't find any evidence in the multiparty docs that the whole upload arrives in a single part, so I'm assuming the library just emits each part with a byte length and offset and leaves any concatenation up to us.

leontastic: One way to know if this works as-is: write a test POSTing a massive file (e.g. 20 MB), then download it with AWS SDK and check if the length/contents match.

We sort of do this with managed uploads, you can use the same technique (although probably should add more logic checking byte contents/length):
https://github.com/jamhall/s3rver/blob/master/test/test.js#L738

lib/controllers.js
```diff
@@ -517,6 +518,88 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+        const form = new multiparty.Form();
+        let chain = Promise.resolve();
+        let fileWasFound = false;
+
+        form.on("error", err => {
+          res.status(400).send(err);
+        });
+
+        form.on("part", part => {
+          if (!part.filename) {
```
leontastic: It took me a while to realized this condition filters out the parts containing the actual file upload. Maybe add comments like `// parse request fields` in this block and `// parse file upload parts` in the `else` block?

lib/controllers.js
```diff
@@ -517,6 +518,88 @@ module.exports = function(rootDirectory, logger, indexDocument, errorDocument) {
...
+                    resolve();
+                  });
+              });
+            });
+          } else {
+            // 'file' must be the last field in the post (this is how aws requires it)
+            fileWasFound = true;
+            chain = chain.then(() => {
+              if (!fields.bucket) {
+                return res.status(405).send(`
```
leontastic: 👍early return

test/test.js
```diff
@@ -187,6 +187,71 @@ describe("S3rver Tests", function() {
...
+        key,
+        file: fs.createReadStream(path.join(__dirname, "resources/image0.jpg"))
+      },
+      resolveWithFullResponse: true
+    });
+    expect(res.statusCode).to.equal(204);
+    const data = yield s3Client
+      .getObject({ Bucket: buckets[0], Key: key })
+      .promise();
+    expect(data.ContentType).to.equal("binary/octet-stream");
```
leontastic: Can you add another expectation checking that the contents match? It would provide an important guarantee that the file is not corrupted during upload – no worries if it's too difficult, just reply in a comment.

test/test.js
```diff
@@ -187,6 +187,71 @@ describe("S3rver Tests", function() {
...
+            path.join(__dirname, "resources/image0.jpg")
+          )
+        },
+        resolveWithFullResponse: true
+      });
+    } catch (err) {
+      error = err;
+      expect(err.response.statusCode).to.equal(400);
+      expect(err.response.body).to.contain(
+        "POST must contain a field named 'key'"
```
leontastic: Can you check for the response strings exactly instead of using `contain`?

test/test.js
```diff
@@ -187,6 +187,71 @@ describe("S3rver Tests", function() {
...
+        formData: {
+          key: "abc"
+        },
+        resolveWithFullResponse: true
+      });
+    } catch (err) {
+      error = err;
+      expect(err.response.statusCode).to.equal(400);
+      expect(err.response.body).to.contain(
+        "POST requires exactly one file upload per request."
```
leontastic: Can you check for the response strings exactly instead of using contain?

Issue #346, opened 2018-11-29
Cors issue


justinmchase:
I'm not quite sure how to configure cors so I can upload files from a browser correctly.

I'm getting this error:
```
Access to fetch at 'http://localhost:4569/example/6jDdcYvGgWJZgDezh' from origin 'http://localhost:8888' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

I'm basically doing a `PUT http://localhost:4569/example/6jDdcYvGgWJZgDezh` from the browser. Here is the `cors.xml` file I have configured on the server:

```
<CORSConfiguration>
  <CORSRule>
    <AllowedOrigin>*</AllowedOrigin>
    <AllowedMethod>GET</AllowedMethod>
    <AllowedMethod>PUT</AllowedMethod>
    <AllowedMethod>POST</AllowedMethod>
    <AllowedMethod>HEAD</AllowedMethod>
    <AllowedMethod>OPTIONS</AllowedMethod>
    <MaxAgeSeconds>3000</MaxAgeSeconds>
    <AllowedHeader>Authorization</AllowedHeader>
  </CORSRule>
</CORSConfiguration>
```

This is purely for local development so, to be honest, if there was a way to just completely enable it for every domain / method, that would be swell.

justinmchase:
It turns out that my issue was that I was lacking `<AllowedHeader>Content-Type</AllowedHeader>`...

The linked PR does fix the issue where you're not specifying allowed headers as might happen if querying from curl but all I had to do to fix my issue was add that allowed header and it worked. Would be nice to have wildcards for these fields like:

```xml
<CORSConfiguration>
  <CORSRule>
    <AllowedOrigin>*</AllowedOrigin>
    <AllowedMethod>*</AllowedMethod>
    <AllowedHeader>*</AllowedHeader>
    <MaxAgeSeconds>3000</MaxAgeSeconds>
  </CORSRule>
</CORSConfiguration>
```

PR #347, opened 2018-11-29
Handle the case where the access-control-request-header is not provided

justinmchase:
This fixes issue #346 

So for whatever reason the client I am using, when making an `OPTIONS` request to `s3rver` is not specifying any `Access-Control-Request-Header` headers at all. As a result `s3rver` erroneously responds with a 403 error. S3 proper does not have this same issue for the exact same request.

I have added a test which fails with vanilla master code and passes once I made the corresponding fix.

The issue appears to be [here](https://github.com/jamhall/s3rver/blob/master/lib/cors.js#L90) where if the header is not specified then `requestHeaders` ends up being `[ '' ]` rather than `[]`. Then when we get to the check [here](https://github.com/jamhall/s3rver/blob/master/lib/cors.js#L100) the conditional expression `allowedHeaders.length < requestHeaders.length` fails because `[].length < [ '' ].length === true`.

Therefore my solution is to remove empty elements from the set since that should never happen provided there is a value specified in the header.

# Before
curl:
```sh
$ curl -I -X OPTIONS -H "access-control-request-method: PUT" -H "origin: localhost:8888" "http://localhost:4569/example/abc123"
HTTP/1.1 403 Forbidden
Content-Type: application/xml; charset=utf-8
Content-Length: 344
ETag: W/"158-dYD/owyTT+IOwBQKGAb8GxwXdSU"
Date: Thu, 29 Nov 2018 21:23:53 GMT
Connection: keep-alive
```
s3rver:
```sh
$ node .
now listening on host 0.0.0.0 and port 4569
info: [S3rver] OPTIONS /example/abc123 403 344 - 3.963 ms
```

# After
curl:
```sh
$ curl -I -X OPTIONS -H "access-control-request-method: PUT" -H "origin: localhost:8888" "http://localhost:4569/example/abc123"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: PUT, POST, DELETE, HEAD, OPTIONS
Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
Date: Thu, 29 Nov 2018 21:25:27 GMT
Connection: keep-alive
Content-Length: 0
```
s3rver:
```sh
$ node .
now listening on host 0.0.0.0 and port 4569
info: [S3rver] OPTIONS /example/abc123 200 - - 1.668 ms
```

### DIFF
--- a/lib/cors.js
+++ b/lib/cors.js
@@ -86,9 +86,10 @@ module.exports = function cors(config) {
         return res.status(403).send(template);
       }
 
-      const requestHeaders = (
-        req.get("access-control-request-headers") || ""
-      ).split(",");
+      const requestHeaders = (req.get("access-control-request-headers") || "")
+        .split(",")
+        .filter(h => h);
+
       const allowedHeaders = matchedRule
         ? requestHeaders
             .map(header => header.trim().toLowerCase())
--- a/lib/cors.js
+++ b/lib/cors.js
@@ -86,9 +86,9 @@ module.exports = function cors(config) {
         return res.status(403).send(template);
       }
 
-      const requestHeaders = (req.get("access-control-request-headers") || "")
-        .split(",")
-        .filter(h => h);
+      const requestHeaders = req.get("access-control-request-headers")
+        ? req.get("access-control-request-headers").split(",")
+        : [];
 
       const allowedHeaders = matchedRule
         ? requestHeaders
--- a/test/test.js
+++ b/test/test.js
@@ -1322,7 +1322,7 @@ describe("S3rver CORS Policy Tests", function() {
         url,
         headers: {
           origin,
-          "Access-Control-Request-Method": "GET",
+          "Access-Control-Request-Method": "GET"
           // No Access-Control-Request-Headers specified...
         }
       });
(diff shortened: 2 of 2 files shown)


lib/cors.js
```diff
@@ -86,9 +86,10 @@ module.exports = function cors(config) {
         return res.status(403).send(template);
       }
 
-      const requestHeaders = (
-        req.get("access-control-request-headers") || ""
-      ).split(",");
+      const requestHeaders = (req.get("access-control-request-headers") || "")
+        .split(",")
+        .filter(h => h);
```
kherock: I think you could make it clearer here that you're (properly) handling the edge case where the `access-control-request-headers` string is empty to begin with. Could you perhaps rewrite this as a ternary
```javascript
req.get('access-control-request-headers')
  ? req.get('access-control-request-headers').split(',')
  : [];
```
similar to how we determine `allowedHeaders`?

justinmchase: No problem, done.

leontastic:
@justinmchase Thank you for your contribution. I just cut a new release with this fix: https://github.com/jamhall/s3rver/releases/tag/v2.2.7

PR #373, opened 2019-01-24
Fix range handling for oversized and out of range situations

jnardone:
Range requests were doing a few things wrong:

1) always returning true regardless of whether or not the range was successful.  If the range is completely out of bounds a 416 must be returned
2) returning the requested range as the size and not the actual amount of data returned

This broke things that depend on range requests (notably s3-blob-store & feathers-blob) which we use extensively.  Hoping for a quick merge & tag so we can keep pulling from the real upstream.  

Thanks for a great project!

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -253,9 +253,18 @@ class FilesystemStore {
         if (err) return done(err.code === "ENOENT" ? null : err);
         const object = new S3Object(bucket, key, content, metadata);
         if (options && (options.start || options.end)) {
+          const bytesRead = content.bytesRead;
+          const totalBytes = metadata["content-length"];
+          if (
+            bytesRead === 0 &&
+            (options && options.start && options.start >= totalBytes)
+          ) {
+            return done("ERANGE");
+          }
+
           object.range = {
             start: options.start || 0,
-            end: options.end || object.size - 1
+            end: options.start || 0 + bytesRead
           };
         }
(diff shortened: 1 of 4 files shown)


kherock:
Thanks for the simple fix, and nice use of [`fs.ReadStream#bytesRead`](https://nodejs.org/api/fs.html#fs_readstream_bytesread). With that, could you update the engineVersion in package.json to be `>= 6.4.0`? LGTM otherwise.

jnardone:
Will do/did.  Engine ver bumped and added an appropriate response header for the 416 response.  Let me know if you need anything else.

jnardone:
Anything else needed for this to get merged?

leontastic:
@jnardone I've cut a release [v2.2.8](https://github.com/jamhall/s3rver/releases/tag/2.2.8) with your fix and published to npm. Thank you for your contribution!

lib/stores/filesystem.js
```diff
@@ -253,9 +253,18 @@ class FilesystemStore {
...
+          if (
+            bytesRead === 0 &&
+            (options && options.start && options.start >= totalBytes)
+          ) {
+            return done("ERANGE", totalBytes);
+          }
+
           object.range = {
             start: options.start || 0,
-            end: options.end || object.size - 1
```
lukiano: I think this broke the range response header in 2.2.8, but looks like it's fixed on master.

PR #384, opened 2019-02-07
reimplement #373 to work on Windows

kherock:
This ensures no file handles are left open after receiving an unsatisfiable range request by never opening the stream in the first place.

Also I found that on my end, the value reported  by `fs.ReadStream#byteRead` was often 0 bytes since I'm guessing that value only reflects what has been lazily read.


leontastic:
Solid refactor @kherock excellent work!

PR #386, opened 2019-02-07
Require at least one Object in deleteObjects requests

kherock:
Due to the way XML works, it's actually impossible to specify and empty array of Objects, so my concerns that an empty array would actually be allowed don't apply.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -605,7 +605,7 @@ describe("S3rver Tests", function() {
     expect(res.headers).to.have.property("content-length", "100");
   });
 
-  it("out of bounds range requests should return 416", function*() {
+  it("should return 416 error for out of bounds range requests", function*() {
     const file = path.join(__dirname, "resources/image0.jpg");
     const filesize = fs.statSync(file).size;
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -1059,10 +1059,17 @@ describe("S3rver Tests", function() {
...
+    let error;
+    try {
+      yield s3Client
+        .deleteObjects({ Bucket: buckets[2], Delete: { Objects: [] } })
+        .promise();
+    } catch (err) {
+      error = err;
+    }
+    expect(error).to.exist;
+    expect(error.code).to.equal("MalformedXML");
```
leontastic: @kherock Do you think this could break users' code? I can imagine a naive `deleteObject` user might make empty requests sometimes. Should this be released in a patch, minor, or major release?

kherock: As of a few patches ago, I'm pretty certain the server would just throw an unhandled error inside that callback function, so bringing this up to spec shouldn't be a big deal. I vote patch release.

PR #387, opened 2019-02-07
Fix empty directory cleanup when deleting a deeply nested key

### LINKED ISSUE #375: Removing last nested object leaves directories behind
It looks like this was raised in #39 and partially fixed in #103.

For example, if a bucket only contains one object with key `dir1/dir2/dir3/file`, calling `deleteObject` leaves behind the directories `dir1/dir2`.  It looks like `utils.removeEmptyDirectories` is recursing top-down through the directories, so it only deems `dir3` to be empty.

Full example:

_index.js_
```js
'use strict';
const {S3} = require('aws-sdk');
const s3 = new S3({ endpoint: 'http://localhost:4568', s3ForcePathStyle: true });

(async () => {
  await s3.putObject({
    Bucket: 'my-bucket',
    Key: 'dir1/dir2/dir3/file',
    Body: 'file 1'
  }).promise();

  await s3.deleteObject({
    Bucket: 'my-bucket',
    Key: 'dir1/dir2/dir3/file'
  }).promise();

  await s3.deleteBucket({ Bucket: 'my-bucket' }).promise();
})();

```
```shell
> mkdir my-bucket
> s3rver -d $PWD &
> node .
(node:11098) UnhandledPromiseRejectionWarning: BucketNotEmpty: The bucket your tried to delete is not empty
> ls -R my-bucket
my-bucket:
dir1

my-bucket/dir1:
dir2

my-bucket/dir1/dir2:
```

kherock:
Resolves #375. The old implementation didn't check if the directory was empty after running the function on its children.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1623,7 +1623,7 @@ describe("Data directory cleanup", function() {
     }
   });
 
-  it("Can delete a bucket that is empty after some key that includes a directory has been deleted", function*() {
+  it("Can delete a bucket that is empty after some key nested in a directory has been deleted", function*() {
     const bucket = "foobars";
 
     let server;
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -1643,20 +1643,18 @@ describe("Data directory cleanup", function() {
...
       yield s3Client.deleteBucket({ Bucket: bucket }).promise();
-    } catch (err) {
-      throw err;
     } finally {
       yield thunkToPromise(done => server.close(done));
     }
   });
 
-  it("Can put an object in a bucket that is empty after some key that does not include a directory has been deleted", function*() {
+  it("Can put an object in a bucket after all objects are deleted", function*() {
```
leontastic: Nice that is a _way better_ test description

PR #391, opened 2019-02-13
Koa rewrite

kherock:
In particular, this focuses on
* improved error handling and error formatting
* making S3 features more modular
* significantly more accurate static website behavior
* simplify implementing new features in the future

This also fixes support for Node.js 10+


kherock:
@leontastic review at your own peril, I know this is a lot. I'm hoping you'll like with the approach I've taken. The only additional tests this should need are website redirects (redirecting /some/key to /some/key/ with a 302 when /some/key/index.html exists) and for XML-formatted 404 pages when making requests with the aws sdk.

Issue #330, opened 2018-10-24
Support for tus.io resumable file uploads


bsingr:
Example project setup, but is not fully working due to missing ListParts implementation:

- https://github.com/bsingr/tus-nodejs-fake-s3-demo
- https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListParts.html

leontastic:
@bsingr Thank you for posting this issue, looks like ListParts is indeed not supported.

leontastic:
Failing test case should use this function from `aws-sdk`: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listParts-property

Issue #110, opened 2018-01-25
Support NextMarker for listObjects and NextContinuationToken for listObjectsV2


leontastic:
We currently don't return NextMarker in truncated List Object V1 responses: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html

Nor do we return NextContinuationTokens in truncated List Object V2 responses: https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

Would be great to support these so users can write integration tests for cases where >1000 objects exist in a bucket.

kherock:
I'm almost done with this, currently trying to RE continuation token generation for fun. I don't have any solid leads, but so far I've discovered
* the base64-encoded token always begins with `1`
  * furthermore, the first 8 bytes as a UInt32BE tend to exist within the 3560000000-3620000000 range
* the length of the token can vary greatly and is correlated with the length of the key (with the bucket name, and potentially region)
* the length of the token increases in 7-9 byte increments for every 8 characters in the key name - from my brief experimentation:

| Key | Key Length | Token Length |
| ---- | ------------ | --------------- |
`us-east-1/<04-char bucket name>/a` | 16 | 24 bytes
`us-east-1/<04-char bucket name>/abcdefgh` | 23 | 24 bytes
`us-east-1/<04-char bucket name>/abcdefgh` `i` | 24 | 33 bytes (+9)
`us-east-1/<04-char bucket name>/abcdefgh` `ijklmnop` | 31 | 33 bytes
`us-east-1/<04-char bucket name>/abcdefgh` `ijklmnop` `q` | 32 | 41 bytes (+8)
`us-east-1/<04-char bucket name>/abcdefgh` `ijklmnop` `qrstuvwx` | 39 | 41 bytes
| | |
`us-east-1/<26-char bucket name>/a` | 38 | 41 bytes
`us-east-1/<26-char bucket name>/ab` | 39 | 41 bytes
`us-east-1/<26-char bucket name>/ab` `c` | 40 | 48 bytes (+7)
`us-east-1/<26-char bucket name>/ab` `cdefghij` `k` | 49 | 57 bytes (+9)
`us-east-1/<26-char bucket name>/ab` `cdefghij` `klmnopqr` `s` | 57 | 65 bytes (+8)
`us-east-1/<26-char bucket name>/ab` `cdefghij` `klmnopqr` `stuvwxyz` `0` | 64 | 72 bytes (+7)

[...]

PR #392, opened 2019-02-13
3.0.0

### LINKED ISSUE #107: Use Koa instead of Express
I've contributed enough to this project to know it decently well and I think it could benefit pretty massively from async/await and Koa's support for processing the response body _after_ middleware is run.

I'm just opening this issue seeking an opinion on potentially including a Babel build pipeline if I were to write for Koa 2. However it looks like Travis is only doing tests for Node >= 6, so Koa 1 would also work fine if we want to avoid Babel. This isn't intended to be just changes for the sake of using newer features; it really should make things a lot more readable and approachable for adding new features such as versioning.

### LINKED ISSUE #115: Index document, Error document, and CORS config should all be per-bucket options
Normally they must be manually enabled/set after creating each bucket in the S3 console. This can cause problems if you want to have one bucket doing static web hosting and another for normal storage/retrieval.

If we also get around to implementing versioning it's also important that it can be enabled per-bucket.

### LINKED ISSUE #153: Support actions via presigned urls
You can create presignedUrls with s3.getSignedUrl method.

A url looks like this: `http://localhost:3001/file-uploads-temporary/de456c2c-0f93-4df7-ad82-0c2943e447f0.jpeg?AWSAccessKeyId=123&Content-Type=image%2Fjpeg&Expires=1519727198&Signature=L0RMhVZCyfrp37sDZbVfSLCuOSU%3D&x-amz-acl=private`

Currently the metadata (aka Content-Type `Content-Type=image%2Fjpeg`) is not added to the `.dummys3_metadata` file.

I would suggest that we check for the `Content-Type` param and if present simply use that as the value for the objects content-type.

The current workaround is to add the header to your request e.g.

```js
"use strict";
const fetch = require("node-fetch");
const S3 = require("aws-sdk");

const s3 = new S3(config);

const s3Params = {
  Bucket: "some-bucket-name",
  Key: "my-file.jpeg",
  ContentType: "image/jpeg",
  ACL: "private"
};

s3
  .getSignedUrl("putObject", s3Params)
  .promise()
  .then(uploadUrl => {
    const readStream = fs.createReadStream("my-local-image.jpeg");
    return fetch(uploadUrl, {
      method: `PUT`,
      body: readStream,
      // On AWS you do not have to specify this. The Header is taken from the presigned url
      headers: {
        "Content-Type": "image/jpeg"
      }
    });
  });

```

What do you think? @specialkk @leontastic?

If you agree I would come up with a PR for this.

### LINKED ISSUE #306: PUT large object doesn't create custom metadata
When I create large objects, the metadata doesn't seem to be created. For example, a 12mb file I transfer to a S3rver instance creates a `.dummys3_metadata` whose contents are `{"contentType":"binary/octet-stream","customMetaData":[]}`. 

Have you observed this issue before?

### LINKED ISSUE #390: BUG & FIX: Windows path bugfix for "Error: ... contains invalid WIN32 path characters."
To bugfix "invalid WIN32 path characters" insert following code.

This error message is produced if you use the Amplify private storage put command to upload on s3erver on windows. The identity id contains an ":" so the path for the private upload is "private/region:identity_id/file" and windows does not allow special characters like ":", "<"...

At:
https://github.com/jamhall/s3rver/blob/9d8e6ca7ea1d18583e7aeaaa766bfd194835bfe6/lib/models/s3-object.js#L6-L9

after `this.key= key;` insert following code:

```
if(process.platform === "win32") {
      this.key = this.key.replace(/[<>:"|?*]/g, function (match, capture) {
        return  "%" + Buffer.from(match, 'utf8').toString("hex");
      });
    }
```

kherock:
Working changelog:
============

Breaking changes:
* [x] You must now specify a specific set of credentials for signed request work. Use `S3RVER` for both the Access Key ID and Secret Access Key.
* [x] The underlying app returned by `new S3rver(opts)` is now a Koa instance. (#391, resolves #107)
  * [x] Additionally, some constructor options have been renamed and moved.
* [x] `s3rver.run([callback])` no longer returns a HTTP server instance. The HTTP server created is set at `s3rver.httpServer`.
* [x] `s3rver.getMiddleware()` was renamed to just `s3rver.middleware()`
* [x] The RxJS Observable `s3rver.s3Event` has been removed in interest of keeping the library lightweight. Use `s3rver.on('event', handler)` instead.
* [x] Bucket features are no longer globally configured. CORS and static website configurations can be set through S3's API or in custom bucket configurations initialized at server startup.

Features:
* [x] You can now have preconfigured buckets created at server startup via `opts.configureBuckets` (or `--configure-bucket` in the CLI). Existing buckets will have their configurations updated with a warning.
* [x] Support the S3 [cors](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTcors.html) and [website](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTwebsite.html) REST APIs
  * [x] This comes with per-bucket configuration of these features (and any others implemented in the future) (#397, resolves #115)
* [x] Support for following redirects specified by `x-amz-website-redirect-location` metadata (#397)
* [x] Stubs for most other bucket actions (they now report 501 Unimplemented)
* [x] Signature verification and presigned URLs (#419)
* [x] Support specifying custom metadata through query parameters (#419, resolves #153)
* [x] Overriding response headers in presigned requests via `response-*` query params (#419)
* [ ] ~~Support ListParts API~~
* [ ] ~~Support specifying a custom storage backend via `opts.store`~~

Fixes:
* [x] Support more regions for buckets specified via subdomain
  * For example, eu-west-1 only offers an endpoint at s3.eu-west-1.amazonaws.com (with a period instead of a hyphen after .s3)
* [x] Buckets operating in "website mode" behave more predictably and should always serve XML-formatted API responses. Real S3 accomplishes this by having totally separate listeners at the DNS/server level, we reconcile this with a single listener by inpecting the Host header and accepted content type.
  * This fixes the "Unexpected token ..." errors reported by the AWS SDK
  * More specifically, S3rver will prioritize Host and fall back to the Accept header. For example, when applicable, requests with  
  `Host: <bucket-name>.s3[.<region>].amazonaws.com`  
  will always have XML-formatted responses, but requests with 
  `Host: <bucket-name>.s3-website.<region>.amazonaws.com`  
  will always have HTML-formatted responses. Otherwise, S3rver will return HTML when `Accept: text/html` is present.
* [x] Virtual Host-style requests (specifying `Host: <bucket-name>`) should work regardless of static website configuration
* [x] Metadata missing from multipart uploads (#115, fixes #306)
* [x] Querying an object for a key stored without a trailing slash would incorrectly match keys stored with one (#115)
* [x] Can't store keys using reserved characters under Windows (#420, fixes #390)
* [x] All error responses reported by S3rver have been revisited and revised to look nearly identical to real S3


kherock:
(btw tests are passing on Node 8, but it looks like Node 10 breaks our app so this release is a bit overdue)

This is working up to be a pretty big release. I do want to also address #110, #153, #306, and #330 before cutting 3.0 since I think those are all pretty essential features. Working with the old code was a pain and didn't leave much room for fixing/adding new features, so I'm not personally up for maintaining the 2.x branch while Node v6 still has a couple months of maintenance before EOL.

@leontastic what are your thoughts on making a new major release without Node 6 support?

philippdonabauer:
@kherock please check my [https://github.com/jamhall/s3rver/issues/390](issue) and make this little changes.

wilson208:
@kherock thanks for your work on this, I find this to be a really useful tool and use it in a couple of different projects currently. I have came across scenario where I need to use pre-signed urls, so seeing this feature go in would be great for next release! 

Thanks

leontastic:
@kherock I am OK without Node 6 support in next major release.

kherock:
@leontastic I'm planning on having everything I want for 3.0 finished by the weekend - in the meantime can we get 2.2.9 published?

jamhall:
@kherock  - Hi! Thank you for all of your hard work! I have added you as a maintainer to s3rver on npm. You should be able to push releases now :-)

leontastic:
@jamhall Thanks!!

@kherock please feel free to publish/merge without my approval, I trust that you'll be responsible!

kherock:
Sorry that this hadn't seen any updates in a while, I got really busy the past few weeks (should be starting a new job next week 🤞). In the meantime I finally got v2.2.9 published.

I don't have an ETA for this release, but I think I might patch out the last outstanding bug (#390) and cut it there. I'll try to have everything else show up in v3.1. Thanks for being patient.

kherock:
3.0.1 is published! The CLI script had a bad path in the initial 3.0.0 release and was fixed for 3.0.1.

PR #397, opened 2019-02-18
Separate bucket configs for CORS and website policies

kherock:
This should be the last major refactor before implementing the rest of the features planned for 3.0

This adds the concept of "subresources" to the store interface. Currently only bucket cors and website configs use this interface, but it's written to generically support other subresources like object/bucket ACLs, tagging, etc.

Since bucket configurations used to be globally applied, this has substantial implications to the CLI and programmatic interface. I've added the `prefabBuckets[]` option (or `--create-bucket` in the CLI) for composing buckets individually at startup that should work as a suitable replacement for users relying on the old `cors`, `indexDocument`, and `errorDocument` options. I've included sample configs in the `example` folder at the project root.

<sub>(This also adds Prettier markdown formatting, I originally meant to include it as a separate commit/PR but it ended up getting lumped into this)</sub>


kherock:
@leontastic This is also ready to be reviewed. BTW, I was having trouble deciding on what to call the new `prefabBuckets` option - I essentially I want to communicate "ensure and configure buckets". I don't like `createBuckets` since it's not a good analog to AWS's actual `createBucket` method for S3. The other name I had in mind was `precomposeBuckets`.

leontastic:
@kherock It might be a while before I'll have time to review this diligently... If you are confident in your changes and are able to devote time to fixing bugs and responding to issues, I say go for it. No need to wait on me – you're probably more familiar with the codebase than I am now.

Thanks for your hard work, I know the community appreciates it. :smile:

PR #402, opened 2019-02-19
Ensure filesystem folder can be accessed under Linux

adam-nielsen:
When running under Linux (and probably Mac), new buckets are created without execute permissions, which makes it difficult to access properly through the filesystem.

This patch includes execute permissions (used to be able to see inside the folder).  These are then limited by the umask set by the user, reflecting the standard behaviour in other applications.


kherock:
Sorry that this got ignored for so long! This has already been fixed upstream in the `next` branch that I intend to release as v3.0 "soon". I'll leave this open until then.

kherock:
This is fixed in v3.0!

PR #421, opened 2019-03-24
handle s3rver launched with ip other than localhost

allan-simon:
(for example docker environment), now we don't only test for ctx.hostname being 127.0.0.1, but any ip
other when using docker, you would end up with an ip in for example 172.12.0.6, and vhost.js would
have erroneously replaced ctx.path by  `/undefined/$bucketname`

### DIFF
--- a/lib/middleware/vhost.js
+++ b/lib/middleware/vhost.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const net = require("net");
+
 /**
  * Middleware that rewrites URLs for buckets specified via subdomain or host header
  */
@@ -15,7 +17,7 @@ module.exports = () =>
       if (bucket) {
         ctx.path = `/${bucket}${ctx.path}`;
       }
-    } else if (ctx.hostname !== "localhost" && ctx.hostname !== "127.0.0.1") {
+    } else if (ctx.hostname !== "localhost" && !net.isIP(ctx.hostname)) {
       // otherwise attempt to distinguish virtual host-style requests
       ctx.path = `/${bucket}${ctx.path}`;
     }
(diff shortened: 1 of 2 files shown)


lib/middleware/vhost.js
```diff
@@ -15,7 +17,7 @@ module.exports = () =>
       if (bucket) {
         ctx.path = `/${bucket}${ctx.path}`;
       }
-    } else if (ctx.hostname !== "localhost" && ctx.hostname !== "127.0.0.1") {
+    } else if (ctx.hostname !== "localhost" && !net.isIP(ctx.hostname)) {
       // otherwise attempt to distinguish virtual host-style requests
       ctx.path = `/${bucket}${ctx.path}`;
```
kherock: Thanks for this! Though this doesn't totally fix the underlying issue. I believe this line should also read

```javascript
  ctx.path = `/${ctx.hostname}${ctx.path}`
```

Regarding tests, I'm realizing the existing test should be named <code>should reach the server with a bucket ~~vhost~~ subdomain</code>. The behavior fixed in this PR should probably add a couple tests for vhost-style and IP-style requests.

kherock:
I was worried my description of the testing suite fixes needed might not have been too clear, so I went ahead and made the changes myself. Again, thanks for pointing this out!

lib/middleware/vhost.js
```diff
@@ -15,9 +17,9 @@ module.exports = () =>
       if (bucket) {
         ctx.path = `/${bucket}${ctx.path}`;
       }
-    } else if (ctx.hostname !== "localhost" && ctx.hostname !== "127.0.0.1") {
+    } else if (!net.isIP(ctx.hostname) && ctx.hostname !== "localhost") {
       // otherwise attempt to distinguish virtual host-style requests
-      ctx.path = `/${bucket}${ctx.path}`;
+      ctx.path = `/${ctx.hostname}${ctx.path}`;
```
allan-simon: hmmm I will retest tomorrow but in my case i was going in that if and it was working fine with `ctx.path = bucket/path`

PR #422, opened 2019-03-24
add method POST /:bucket to post a new object

allanlegalstart:


allanlegalstart:
if I run `npm run prettier-fix` it actually changes most of the files :/

allan-simon:
@jamhall  I'm interested in a bit of help for the code-style failure , as I'm not that familiar with js/node-js code convention

lib/stores/filesystem.js
```diff
@@ -245,6 +246,22 @@ class FilesystemStore {
     }
   }
 
+  async postObject(object) {
```
kherock: This function isn't needed, you can just use the store's `putObject` implementation. S3's POST endpoint is essentially just an HTML form-friendly PUT endpoint.

lib/routes.js
```diff
@@ -103,6 +103,8 @@ router
     switch (ctx.params.queryMethod) {
       case "delete":
         return objectCtrl.deleteMultipleObjects(ctx);
+      case undefined:
+        return objectCtrl.postObject(ctx);
```
kherock: Could you make this consistent with the other route handlers and keep the `undefined` case at the top of the switch statement?

lib/controllers/object.js
```diff
@@ -225,6 +226,62 @@ exports.getObjectAcl = async function getObjectAcl(ctx) {
   };
 };
 
+
+/**
+ * POST Object
+ * Post an object into a bucket
```
kherock: This could use a better description - the first sentence in the link for S3's API docs usually provides a good summary if you're feeling lazy.

lib/controllers/object.js
```diff
@@ -225,6 +226,62 @@ exports.getObjectAcl = async function getObjectAcl(ctx) {
...
+    ctx.logger.error(
+      'Error uploading object to bucket "%s"',
+      ctx.params.bucket,
+      err
+    );
+    throw err;
+  }
+};
+
+const _promisifyUpload = (req) => new Promise((resolve, reject) => {
```
kherock: You can just have this function inlined.

allanlegalstart:
for prettifier-fix, it actually changes most of files, not just terminal output 

i.e example of changes I see after running it:

```
iff --git a/README.md b/README.md                                                                                                                                                                                                           
index 53e8672..61eb65f 100644
--- a/README.md
+++ b/README.md
@@ -121,17 +121,17 @@ Creates a S3rver instance
 For your convenience, we've provided sample bucket configurations you can access using `require.resolve`:
 
 javascript
-const corsConfig = require.resolve("s3rver/example/cors.xml");
-const websiteConfig = require.resolve("s3rver/example/website.xml");
+const corsConfig = require.resolve('s3rver/example/cors.xml')
+const websiteConfig = require.resolve('s3rver/example/website.xml')
 
 const s3rver = new S3rver({
   prefabBuckets: [
     {
-      name: "test-bucket",
-      configs: [fs.readFileSync(corsConfig), fs.readFileSync(websiteConfig)]
-    }
-  ]
-});
+      name: 'test-bucket',
+      configs: [fs.readFileSync(corsConfig), fs.readFileSync(websiteConfig)],
+    },
+  ],
+})



```

and t

lib/controllers/object.js
allanlegalstart: done

lib/controllers/object.js
allanlegalstart: done

lib/routes.js
allanlegalstart: done

lib/stores/filesystem.js
allanlegalstart: I'm not sure how true it is, I'm not a nodejs developer so I may be wrong,  but the big difference is that in my  case the file is already in the filesystem, so here we merely move it + hash it from FS, while the put implementation take the data from the response's body, so I'm a bit clueless on how to torture the controller to make it works with the existing filesyste/putObject

kherock: OK I think I see - when it's defined, `object.content` is expected to be a stream. I was under the impression multiparty would expose streams for any files it discovered.

I think you should be able to call `fs.createReadStream(files['file'][0].path)` when you construct the `S3Object` in the controller.

kherock:
The prettier issues you're encountering are strange. The snippet you posted is odd since it's removing semicolons and adding trailing commas. Is there a `.prettierrc` file somewhere in a parent directory?

allanlegalstart:
yes you were right, my company's project had a prettierc file at the root, now should be good for the style

allanlegalstart:
for the test, I haven't found in the aws js sdk , a  "postObject" method that I could use to mimic the test you've made for the putObject

kherock:
This initial implementation is pretty simple, so I think just one tests case using `request` to construct a POST request should be fine. POST behavior can get pretty complicated once you start factoring in signature validation and POST policies.

[Here's an example](https://github.com/request/request#multipartform-data-multipart-form-uploads) of making a multipart upload using `request`. Keep in mind that the `request` implementation we use in tests supports Promises instead of callbacks.

allanlegalstart:
ok thanks for the pointer I will try to write a test tonight

allan-simon:
@kherock here's your test :) please tell me everything is ok ^^'

allan-simon:
@kherock  any chance for this to get a second review ? :/

kherock:
There are some relatively minor things I'd like to see addressed (the async code style is still off, and using intermediate storage for writing files from the multipart request shouldn't be necessary). But those are things I think I'll resolve on my own when I have time.

I've rebased your branch for you (the commit log was starting to run long). Thanks for your patience.

allan-simon:
ok :( to be honest I'm no javascript developer so I don't know how to correct these stuff :(

allan-simon:
oh I didn't see it was merged, nice \o/

kherock:
@allan-simon no worries, this is definitely a good enough start. I wanted to address a few other issues that have been in my backlog for this weekend, this feature should be included in v3.5.0!

allan-simon:
nice, I will rebase https://github.com/jamhall/s3rver/pull/505 today then, because that's the last bug preventing us from using the official release

PR #424, opened 2019-03-25
Fully implement List Objects V2 with pagination response elements

### LINKED ISSUE #110: Support NextMarker for listObjects and NextContinuationToken for listObjectsV2

kherock:
Resolves #110. This implements "starts-after" for GET Bucket v2 as well all pagination-related response data for both the v1 and v2 APIs. This also fixes the behavior for "marker"/"starts-after" (they erroneously require a key starting with the marker to exist in order to work properly).

I also ended up doing some significant refactoring to the tests around listObjects since they weren't very strict and also ran very slowly (it actually isn't necessary to generate 1000+ objects to make adequate test cases most of the time).

### DIFF
--- a/lib/controllers/bucket.js
+++ b/lib/controllers/bucket.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const { DUMMY_ACCOUNT } = require("../models/account");
 const S3Error = require("../models/error");
 const {
   S3CorsConfiguration,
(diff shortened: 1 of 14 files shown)


leontastic:
@kherock Heroic effort man keep it up 💪

PR #455, opened 2019-05-12
Add md5 extension to md5 file

mtmckenna:
Hello! This issue came up while trying to copy objects.

I think there's an issue with the `getMetadata` function overwriting the object file instead of the md5 file. Adding the md5 extension as in this PR seemed to do the trick.

Please let me know if I can provide any more info! Thank you!

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -334,7 +334,7 @@ class FilesystemStore {
             this.getResourcePath(srcBucket, srcKey, "object.md5"),
             this.getResourcePath(destBucket, destKey, "object.md5")
           )
-        ])
+        ]);
       }
       return this.getMetadata(destBucket, destKey);
     }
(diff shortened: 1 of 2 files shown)


kherock:
Thanks for the PR! Normally those md5 files are generated upon file upload, and that codepath should only happen if they were somehow deleted by the next time the object is read. Can you think of any reason why those files wouldn't be generated? Just curious.

mtmckenna:
Hello, @kherock ! I saw this behavior when doing a call to `s3.copyObject` from my code.

Doing the above in the `aws-sdk` eventually hits `copyObject` in `filesystem.js`. Depending on the conditional,`copyObject` calls `getMetadata`, which is what triggers the issue I'm seeing (the md5 file overwriting the object file).

Maybe another way to do this would be to make sure the md5 file is always copied during `copyObject` calls? Not sure how that would impact other things though.

I hope that helps--please let me know if I can provide any more info.

Thank you!

kherock:
That's really helpful - I was just looking for a way to capture this issue in a test case if possible since I was surprised this came up during normal use. It looks like the existing tests don't actually verify that the copied content matches. I'll aim to release a patch tonight!

mtmckenna:
That’s awesome! Your solution is much cleaner. Thank you for your help on this!

kherock:
v3.1.0 is published, thanks again for catching this!

PR #473, opened 2019-06-04
Disable signature function

jpike88:


jpike88:
@kherock I've spent way more time than I wanted to trying to make this work, so I'm just going to add the ability to turn off signature checking altogether.

My staging server is still having issues despite my last PR and best efforts, I'll leave the other issue open for your perusal. If you want to put a fix out, let me know and I'll try it.

jpike88:
also, your tests randomly break by the looks of it...

kherock:
It's likely because of the runaway Promise you now have in the `.callback()` - the bucket configurations that were supposed to be set for the last test are only taking effect by the time the next test runs.

jpike88:
Ah ok I'll fix it

jpike88:
Looks like this wasn't enough to fix the issue, likely due to the path issue going beyond just how the signature is calculated... im closing this and reverting to a fork of the older version sorry, as I said I'll try any fix attempts you have.

If you get stuck, I'll do a hail mary and fire up a cloud9 instance and invite you into it for repro.

It's really remote server using domain name + using middleware in express subpath, wouldn't be hard to fire something up for that. Let me know if you need help

PR #479, opened 2019-06-10
Correct expirestime calculation

SangMo93:


SangMo93:
anybody check something? or Feedback?

SangMo93:
if do not change calculation. presigned url immediately expires when I request presigned url.
expires time and request time are same! so you don't access presigend url

kherock:
Hi, thanks so much for the PR! Sorry I wasn't able to address this before I went on vacation. This looks good as-is, I'll try to publish a patch release this weekend.

kherock:
Thanks for your patience, this is now released in v3.3.0.

PR #503, opened 2019-08-05
Object Tagging support

jagregory:
Still a work-in-progress.

Adds Object Tagging support. PutObjectTagging saves a `key.tagging.xml` subresource in the data directory containing the tags, and GetObjectTagging reads that file.

### DIFF
--- a/lib/controllers/object.js
+++ b/lib/controllers/object.js
@@ -8,7 +8,8 @@ const { DUMMY_ACCOUNT } = require("../models/account");
 const S3Error = require("../models/error");
 const S3Event = require("../models/event");
 const S3Object = require("../models/object");
-const { capitalizeHeader } = require("../utils");
+const { TaggingConfiguration } = require("../models/config");
+const { capitalizeHeader, utf8BodyParser } = require("../utils");
 
 async function xmlBodyParser(ctx) {
   const { req } = ctx;
@@ -229,26 +230,6 @@ exports.getObjectAcl = async function getObjectAcl(ctx) {
   };
 };
 
-exports.getObjectTagging = async function getObjectTagging(ctx) {
-  const key = ctx.params.key;
-  const exists = await ctx.store.existsObject(ctx.params.bucket, key);
-  if (!exists) {
-    throw new S3Error("NoSuchKey", "The specified key does not exist.", {
-      Key: key
-    });
-  }
-
-  const tagging = await ctx.store.getObjectTagging(ctx.params.bucket, key);
-
-  ctx.body = {
-    Tagging: {
-      "@": { xmlns: "http://doc.s3.amazonaws.com/2006-03-01/" },
-      ...tagging
-    }
-  };
-  console.log(ctx.body);
-};
-
 /**
  * PUT Object
  * This implementation of the PUT operation adds an object to a bucket.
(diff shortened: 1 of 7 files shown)


lib/stores/filesystem.js
```diff
@@ -305,6 +326,35 @@ class FilesystemStore {
     return { size, md5 };
   }
 
+  async putObjectTagging(bucket, key, content) {
+    const taggingPath = this.getResourcePath(bucket, key, "tagging.json");
+    const parsedTagging = await parseXmlBody(content);
```
jagregory: @kherock Not sure what's the preferred approach here. The SDK sends tags as XML in the request body, I parse this and then save it to JSON, and then on the way out I do the reverse. I'm wondering if I should just take the payload, save it, and then send it back out again for Get requests without touching it at all?

kherock: That's the approach I would have tried first - you should try seeing if you can reuse `pubSubresource` and `getSubresource` for this. However, those methods are defined to use a config model (such as ones we have for website and CORS configurations) that would parse and validate tags. Let me know if you need more guidance on that.

jagregory:
@kherock I redid this using Subresources and it's a lot simpler, thanks! I reckon this is good for a review when you have time.

lib/controllers/object.js
```diff
@@ -354,6 +355,57 @@ exports.putObjectCopy = async function putObjectCopy(ctx) {
...
+    ctx.logger.error(
+      'Error tagging object "%s" in bucket "%s"',
+      key,
+      bucket,
+      err
+    );
+    throw err;
+  }
+};
+
```
kherock: Could you insert these definitions in the order as they appear in this list here?
https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectOps.html

The structure under `controllers/` is meant to mirror the official REST API docs.

lib/utils.js
```diff
@@ -120,3 +120,19 @@ exports.parseDate = function(dateString) {
 exports.toISO8601String = function(date) {
   return new Date(date).toISOString().replace(/[-:]|\.\d+/g, "");
 };
+
+/**
+ * Reads a request body stream to a string.
+ *
+ * @param ctx
+ * @returns {Promise<string>}
```
kherock: This is actually `Promise<void>`, we're not returning anything. You can also give `ctx` a type of `Koa.Context` if you'd like (there's no TS compatibility yet so it doesn't matter too much to me).

lib/models/config.js
```diff
@@ -341,3 +343,26 @@ class S3WebsiteConfiguration extends S3ConfigBase {
...
+    }
+
+    return new TaggingConfiguration(xml);
+  }
+
+  constructor(config) {
+    super("tagging", config);
+  }
+}
+TaggingConfiguration.empty = new TaggingConfiguration(
```
kherock: How about naming this `TaggingConfiguration.EMPTY`?

jagregory:
Changes pushed :)

I've also added a couple more tests for error scenarios too.

jagregory:
Hey @kherock, anything else needed to merge this? Cheers

kherock:
This is released in v3.4.0! Sorry for the wait.

jagregory:
All good, thanks! 🙂

PR #505, opened 2019-08-07
fix #504, as reponse can both be XML and HMTL documents, we use he.es…

### LINKED ISSUE #504: bug: the url in the <Location> reponse is xml encoded whereas in aws s3 it's urlencoded
i.e aws s3 answers 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<PostResponse><Location>https://foobar.s3.amazonaws.com/pr%C3%A9sident.jpg</Location><Bucket>foobar</Bucket><Key>président.jpg</Key><ETag>"14dd4667441ada2b3914e5c1b72573b3"</ETag></PostResponse>
```

where s3rver answers

```xml
<?xml version="1.0" encoding="UTF-8"?>
<PostResponse><Etag>14dd4667441ada2b3914e5c1b72573b3</Etag><Bucket>foobar</Bucket><Key>pr&eacute;sident.jpg</Key><Location>http://172.17.0.6/foobar/pr&eacute;sident.jpg</Location></PostResponse>
```

i.e  `pr&eacute;sident.jpg` vs `pr%C3%A9sident.jpg`

this is problematic because in Firefox , if you try to put the answers of s3rver in DOMParser, it will fail with a `unknown entity` parsing error, as it does not understand `&eacute;`

allan-simon:
…cape instead of he.encode

as he.escape will produce valid XML while he.encode produce also html entities not known by pure XML

### DIFF
--- a/lib/s3rver.js
+++ b/lib/s3rver.js
@@ -45,8 +45,7 @@ class S3rver extends Koa {
       const parser = new xmlParser.j2xParser({
         ignoreAttributes: false,
         attrNodeName: '@',
-        tagValueProcessor: a =>
-          he.encode(a.toString(), { useNamedReferences: true }),
+        tagValueProcessor: a => he.escape(a.toString()),
       });
       this.use(async (ctx, next) => {
         await next();


allan-simon:
@jamhall  I've rebased and fixed the styling issue :)

PR #522, opened 2019-08-30
Detect SDK/CLI usage with x-amz-* headers and query string parameters

### LINKED ISSUE #502: Requests for images and other non-html content via a website endpoint not handled
The website middleware doesn't handle any requests that don't preference `text/html` in their `Accept` header, which means requests for objects in the bucket that aren't HTML fail.

For example, when a browser makes a request for an image from an `img` element it uses a different `Accept` header than if you open the URL directly (same applies for `video` etc...). MDN has a list of default Accept headers here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values

In short, the current implementation fails to correctly respond to requests for images/videos/other content when the browser initiates the request not the user.

There's a failing test in my fork (branch: website-images): https://github.com/jagregory/s3rver/tree/website-images

jagregory:
Fixes #502

All the tests pass, and I've also tested it locally with my web app.

### DIFF
--- a/lib/middleware/website.js
+++ b/lib/middleware/website.js
@@ -39,14 +39,19 @@ exports = module.exports = () =>
     } else if (
       ctx.hostname.match(/(^|\.)s3([-.].+)?\.amazonaws\.com$/) ||
       !ctx.params.bucket ||
-      // NOTE: This allows web browsers to experience static website behavior while maintaining
-      // SDK compatibility. Real S3 doesn't care about the accepted content or bucket config.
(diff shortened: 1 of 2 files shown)


jagregory:
Just pushed up a formatting fix. I was still running `npm run prettier-fix` which obviously wasn't doing anything :)

Don't stress about the release. I'm just happy I was able to find a fix. I'm running off a fork so no rush.

PR #526, opened 2019-09-03
URL encode the uri part of the stringToSign

### LINKED ISSUE #513: SignatureDoesNotMatch error when signing a request containing unicode characters
When signing a request containing a "é" character for example, I end up getting a 403 Response with a SignatureDoesNotMatch error inside.

This is due to the fact that when creating the stringToSign in S3rver, the path is not URL-encoded.

It has to be URL encoded as per the AWS documentation:
https://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationExamples-8

I am making a PR to fix that.

rafou:
Fix #513

Tested locally. Now we are able to verify signature coming from Python boto with non ascii character (such a "é").

### DIFF
--- a/lib/middleware/authentication.js
+++ b/lib/middleware/authentication.js
@@ -466,7 +466,7 @@ function getStringToSign(canonicalRequest) {
     ...canonicalRequest.amzHeaders,
     canonicalRequest.querystring
       ? `${canonicalRequest.uri}?${canonicalRequest.querystring}`
-      : canonicalRequest.uri,
+      : encodeURI(canonicalRequest.uri),
   ].join('\n');
 }
 


allan-simon:
@jamhall  is this PR missing something to be merged ? (sorry to pushy :/ )

kherock:
@allan-simon no worries - I should also probably advertise myself as the active maintainer of this project somewhere more visible.

Anyway, apologies for the short attention span on this project lately. I'd like to see a test case on this before merging. I'll try to spend some time today adding one in addition to addressing the other open PRs.

allan-simon:
@rafou  should be able to get one, maybe not as a "code test" but may be as a "try this at home" test ? (would it be enough for you ?)

allan-simon:
@kherock  up ,  if we give you repeatable step to reproduce the issue, would it be enough for you to merge this PR ? as we're hitting once again a wall , if we can do anything to help you , let us know :)

kherock:
@rafou sorry for the unnecessarily long delay on such a simple fix. I'll be including this in the next release which will should be out sometime this week!

kherock:
v2 signatures actually had much bigger problems which I've addressed in #630. For this particular issue, AWS actually uses [RFC 3986](https://tools.ietf.org/html/rfc3986)'s reserved char list, so `encodeURIComponent` isn't enough by itself.

rafou:
Thanks @kherock for the explanation and for taking a look at the problem !

Cheers

PR #536, opened 2019-09-13
Make walk over dir tree async

ad-m:
In the case of NodeJS 10 the change could be limited to adding "async", "await" and removing "Sync" in appropriate places. However, the project accepts Node 8 compliance, so it proposes this approach.

The function is used in a fairly critical place and is recursive - file listing in bucket, so it can potentially be called frequently.


ad-m:
Is there anything what I can do to help merge that PR?

kherock:
I'd very strongly prefer waiting until the next major release to make big changes here since the fix is so simple in Node 10+. In the meantime, a simple thing I can probably do to keep listObjects from blocking the server is to just call `await promsifiy(setImmediate)()` at the end of loop each iteration.

PR #537, opened 2019-09-13
Use transform & pipe to manage streams

ad-m:
It allows use NodeJS backpressure: https://nodejs.org/es/docs/guides/backpressuring-in-streams/
In this way we will take out the return value of the .write() function.

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -2,6 +2,7 @@
 
 const crypto = require('crypto');
 const fs = require('fs-extra');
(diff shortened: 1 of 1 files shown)


ad-m:
Is there anything what I can do to help merge that PR?

kherock:
Hi @ad-m, thanks for the contribution this is definitely the right way to do it! I'll admit that until recently I barely knew anything about Transform streams, so I didn't place much priority on this and I'm sorry for the silence. This will be in the next minor release!

Issue #557, opened 2019-11-07
Add the possibility to disable the vhost resolution


ksnll:
In our organisation current testing setup, we don't use `localhost` to access local services but we use another hostname such as `local.organisation`

In newer version of s3rver, using this hostname to access a bucket, will force a resolution to a specific bucket `local.organisation:port/local.organisation`.

For this reason, we cannot update s3rver to a newer version 😢 

It would be nice to have the option to disable the vhost resolution through an option. 
I gave it a go in creating a PR for this

allan-simon:
we have the same problem, we s3rver in our CI, where we can give it one and only one docker alias , which make  bucket name vhost resolution impossible , instead we've always been using the "bucket name in the path" method 

and as the author, that has forced us to some woraround , we instead use the ip adress directly (which may be useful for you @ksnll  )

kherock:
@ksnll This is definitely valid, and I'm planning for a relatively simple workaround in the next minor release where both `os.hostname()` and the custom service endpoint will resolve to the S3 service root. So both of the following should work once it's implemented:

On an OS with hostname `localhost.organisation`, just running
```console
$ s3rver -d ./data
```
would have s3rver resolve the path to `my-bucket` in `http://local.organisation/my-bucket`.

Alternatively, an OS with hostname `s3.localhost.organisation` you could run
```console
$ s3rver -d ./data --service-endpoint local.organisation
```
and s3rver would resolve `my-bucket` for `http://s3.local.organisation/my-bucket/`.

eliasdorneles:
@kherock and what are your thoughts on the solution implemented at: https://github.com/jamhall/s3rver/pull/558 (basically, adding a feature flag) ?

kherock:
I'm still debating what the best behavior will be going forward since path-style access will be slowly disappearing and it's something I want to have addressed for v4. I think regardless people are going to want a way to continue support for path style-access so I'm supportive of something like a `--no-vhost` option.

eliasdorneles:
Yeah, that would be really cool indeed!

And considering it doesn't seem to cost much (since it's already implemented for `localhost` and access with IP address), I see no downsides. :)

eliasdorneles:
That's awesome, thank you!

PR #558, opened 2019-11-07
Make it possible to disable vhost resolution

ksnll:
This PR will add a flag to disable the vhost bucket name resolution.

More information could be found [here](https://github.com/jamhall/s3rver/issues/557)

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ mysite.local 127.0.0.1
 
 Now you can access the served content at `http://mysite.local:4568/`
 
+This behaviour can be disabled by setting the `resolveVhosts` option to `false`.
+
 ## Tests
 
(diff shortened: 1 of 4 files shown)


eliasdorneles:
Thanks, I was hit today by the very same problem described on ticket #557 which this PR fixes.

Maintainers: hello, any chance this PR will move forward?

I understand that [Amazon intends to deprecate the path-style of addressing for buckets created after September 2020](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) but this will still be supported for existing buckets, it would be nice to support that if it doesn't cost much.

We're workarounding this by using a bucket name with the same name as the host name.

allan-simon:
@kherock this PR is quite blocking on our side too,  anything we can do to help ?

PR #571, opened 2019-12-11
Consistently return `this` from close

Retsam:
The JSDoc on `close` claims that it returns` this` when called with a callback, but in practice, it was only returning `this` if the server wasn't running.


kherock:
Thanks for the contribution @Retsam! This should be available in the next release soon.

PR #574, opened 2019-12-20
Properly handle redirects for urls with anchors

DaKaZ:
I recently ran into an issue where I was using s3rver to host an SPA that has routing.  I found this [ingenious hack](https://viastudio.com/hosting-a-reactjs-app-with-routing-on-aws-s3/) to make everything work on S3 but alas it did not work with serverless-offline and the s3rver.  This PR modifies the website routing logic to properly handle anchor tags.

### DIFF
--- a/lib/middleware/website.js
+++ b/lib/middleware/website.js
@@ -72,13 +72,21 @@ exports = module.exports = () =>
       if (err.code !== 'NoSuchKey') throw err;
 
       const key = err.detail.Key;
+      // SPA websites may have url anchors (ie: http://localhost/#my-anchor)
+      const [keyWithoutHash, keyHash] = key.split('#');
 
-      if (config.routingRules) {
+      if (keyWithoutHash === '' || keyWithoutHash.endsWith('/')) {
+        ctx.params = {
+          ...ctx.params,
+          key: keyWithoutHash + config.indexDocumentSuffix + (keyHash || ''),
+        };
+        await getObject(ctx);
+      } else if (config.routingRules) {
         for (const routingRule of config.routingRules) {
           // Only 404s are supported for RoutingRules right now, this may be a deviation from S3 behaviour but we don't
           // have a reproduction of a scenario where S3 does a redirect on a status code other than 404. If you're
           // reading this comment and you have a use-case, please raise an issue with details of your scenario. Thanks!
-          if (routingRule.shouldRedirect(key, 404)) {
+          if (routingRule.shouldRedirect(keyWithoutHash, 404)) {
             const location = routingRule.getRedirectLocation(key, {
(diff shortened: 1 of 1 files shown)


DaKaZ:
@kherock any change we can get this merged?

fishcharlie (changes requested):
Seems like we should add tests here to ensure there is no regression and that this works as expected.

lib/middleware/website.js
```diff
@@ -72,13 +72,21 @@ exports = module.exports = () =>
       if (err.code !== 'NoSuchKey') throw err;
 
       const key = err.detail.Key;
+      // SPA websites may have url anchors (ie: http://localhost/#my-anchor)
+      const [keyWithoutHash, keyHash] = key.split('#');
```
kherock, line 76: I don't like that this mishandles cases where there are `#` characters in the fragment itself (`/some-key##this-is-weird-but-i-dont-judge`). Assigning a `const fragmentIdx = key.indexOf('#')` and using string slices is probably better.

kherock (changes requested):
The approach here unfortunately seems a bit off and I'm not convinced it works yet, this definitely needs tests.

I'm actually not sure if there's any control flow that needs updated to recreate the routing workaround you linked. Based on your implementation, I think there might actually a be a bug in the call to `routingRule.shouldRedirect`? You've actually changed the behavior here so that routing rules are ignored for keys ending with `/` which definitely isn't right.

Now there is a secondary issue you've brought up in that if you configure your S3 website bucket with a custom `ErrorDocument.Key` containing a URL fragment, our middleware will look up the key as if it were sent URI-encoded, attempting to read from a key with a literal `#` in the name. In most cases this will result in a NoSuchKey error. However, this is the only scenario where I could see the implementation here having any effect.

If you're not seeing this behavior on your real S3 setup, the correct behavior might be to strip the fragment when reading the object containg the HTML error page, but that's beyond the scope of the original issue.

leontastic:
Looking at the original article I don't think it's actually necessary to add a hash to the end of the document to accomplish SPA deployment with correct routing. Alternative solution: https://stackoverflow.com/questions/57763686/cheapest-way-to-host-a-spa-or-static-website-with-https-on-s3

However if hashbangs in URLs don't work then this is a bug. A failing test case would be great so we know what behaviour we should expect.

kherock:
I've verified that the example config given works locally in #690 and I've added some test coverage for the problem.

I was also able to verify that specifying an error document key containing a `#` does cause S3 to attempt looking up a that exact key, `#` and all. So in this case S3rver's behavior is correct and there doesn't need to be any special fragment-parsing logic within the website middleware.

Issue #534, opened 2019-09-13
Support Signature Version 4 streaming API


ad-m:
Hello,

I would like notice that all AWS migrated to signature v4. It mean some client eg. Restic no longer able to s3rver. In case of signed chunked data ( https://docs.aws.amazon.com/en_pv/AmazonS3/latest/API/sigv4-streaming#sigv4-chunked-upload-sig-calculation-chunk0 ) saved data is corrupted. 

Here is implementation of Signature v4 in NodeJS which can be inspiring:

- https://github.com/noobaa/noobaa-core/commit/6c95ba38b958d32e24493a4a589abfe630aa6056 (I tested that stream transform and it correctly drop signatures from stream)
- https://github.com/scality/cloudserver/blob/9615d51c9e0500db37905dcfd04b17187132eddf/lib/auth/streamingV4/V4Transform.js

Greetings,

sgillespie:
This is indeed breaking the Java SDK v2, and results in the following error on `putObject`:

```
Exception in thread "main" software.amazon.awssdk.core.exception.SdkClientException: Data read has a different checksum than expected. Was 0xaaf62ecf2a67b8cc3d05b40f04ddb4ff, but expected 0xc
964993c6bbd449be0bca24e4747c053                                                                                                                                                                
        at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:98)                                                                              
        <-- Snip -->
        at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:55)
        at software.amazon.awssdk.services.s3.DefaultS3Client.putObject(DefaultS3Client.java:7119)
```

The workaround I have found is to disable this entirely (syntax in Kotlin):

```
val config = AwsRequestOverrideConfiguration.builder().signer(NoOpSigner()).build()
val req = PutObjectRequest
            .builder()
            .overrideConfiguration(config)
            // other options
            .build()
```

I'd be happy to put together a PR if I can find the time

kherock:
Signature Version 4 support is planned for the next major release, so stay tuned over the next few weeks for that.

However, I haven't really investigated chunked uploads so it might take some time to support them adequately. @ad-m thanks for the reference links!

@sgillespie are the snippets you pasted using `EventStreamAws4Signer`? Let me know if you still have issues using `AwsS3V4Signer`. - I'm just looking for confirmation that you're experiencing the same issue.

[...]

Issue #498, opened 2019-08-01
Credentials can't be configured


c24w:
> [You must now specify a specific set of credentials for signed request work. Use S3RVER for both the Access Key ID and Secret Access Key.](https://github.com/jamhall/s3rver/releases/tag/v3.0.1)

This is fine, but we're using other AWS services in the same app.  This means we'd have to configure the AWS SDK to use different credentials only when using S3, creating test-specific implementation.

We'd prefer to only configure the AWS SDK once with the same test credentials; It would be handy to make the _dummy_ credentials configurable when starting s3rver.

adam-nielsen:
Ah, just wondered why everything broke when I updated `serverless-s3-local`.  Definitely can't change the credentials because we are using them to read data from live databases and want to write to the local S3 instance.

Is there some way we can go back to the old behaviour where the credentials are just ignored?

adam-nielsen:
FYI for anyone else having this problem, downgrading with `npm install serverless-s3-local@0.3.23` dropped me back to s3rver 2.2.5 which got everything working again.

fishcharlie:
I believe this will be fixed in #598? Is that what you are looking for or something else?

c24w:
Yep, that looks about right! :)

scooper91:
@kherock @leontastic Is there any way we can get this issue moving again? It's something we're struggling with at the moment, as we need to talk to this, and a real AWS service in the same application (for local testing).

There seem to be two open pull requests related to it - #598 and #682.

PR #598, opened 2020-03-03
allow custom credentials through environment variables

mlintern:


lib/models/account.js
```diff
@@ -21,6 +21,9 @@ AWSAccount.registry = new Map();
 
 exports = module.exports = AWSAccount;
 
+const accessKeyId = process.env.S3RVER_ACCESS_KEY_ID || 'S3RVER';
+const secretAccessKey = process.env.S3RVER_SECRET_ACCESS_KEY || 'S3RVER';
```
fishcharlie, line 25: It looks like this is being used in 2 different places. Maybe we should move these variables out into a separate file and have a single location to make the code more DRY?

OR have the version in `lib/controllers/bucket.js` reference this one.

fishcharlie (approved):
One suggestion. Not sure it should hold up the PR tho.

lib/controllers/bucket.js
```diff
@@ -28,7 +29,7 @@ function generateContinuationToken(bucket, keyName, region) {
 function decipherContinuationToken(token) {
   const buf = Buffer.from(token, 'base64');
   if (buf.length < 8) return '';
-  const key = Buffer.alloc(8, 'S3RVER', 'utf8');
+  const key = Buffer.alloc(8, secretAccessKey, 'utf8');
```
leontastic, line 32: I think this can simply be a dummy string as long as it matches the reference above on line 15, since the token is not meant to be decoded by anyone but the S3 server. That would eliminate the reference to `secretAccessKey` here.

leontastic (commented):
This is the right approach.

If we're going to support configurable credentials we should probably initialize them in the `S3rver` constructor instead. That way we can support `accessKeyId` and `secretAccessKey` directly in the CLI and API without referencing `process.env`, and leave it up to the user how they want to load their keys (environment variables, via CLI, from file, etc.):

```js
new S3rver({
  // other config options
  defaultAccessKeyId: 'MY_ACCESS_KEY_ID',
  defaultSecretAccessKey: 'MY_SECRET_ACCESS_KEY'
})
```

```bash
$ s3rver --directory s3temp --defaultAccessKeyId "MY_ACCESS_KEY_ID" --defaultSecretAccessKey "MY_SECRET_ACCESS_KEY"
```

lib/controllers/bucket.js
kherock: Yeah this function is definitely missing some documentation explaining what's going on, my bad.

Basically, there's evidence that S3 uses a symmetric key to generate continuation tokens since the token length is proportional to the combined region, bucket name, and object key. The string I used here is arbitrary. If you'd like to have something more descriptive there instead, `'S3RVER_CTOKEN_KEY'` would probably be fine.

kherock:
@leontastic would it be too magical if we inferred default values for those based on credentials loaded in the environment (i.e. ask `aws-sdk` for its currently loaded credentials)?

I did intentionally leave out configurability in the constructor in the original dummy account implementation because to me an S3 API with the concept of "default account" felt like technical debt for adding proper bucket ownership later. Cross-account scenarios are fairly common, but I'm also aware in that case that you're probably a large company who has the resources to easily run those tests against the real thing anyway.

leontastic:
@kherock I don't think it's too magical, but it would force users of the JavaScript API to use multiple configuration methods. Imagine using `dotenv` to load environment variables and instantiating your S3rver instance with JS:

```js
require('dotenv').config()

new S3rver({
  directory: process.env.TMP_S3_DIRECTORY,
  // ... more config
})
```

With the proposed change the only way you could add a default key pair is either baking it into `process.env` (configuration by side-effect):

```js
require('dotenv').config()

process.env.S3RVER_ACCESS_KEY_ID = process.env.AWS_ACCESS_KEY_ID

new S3rver({
  directory: process.env.TMP_S3_DIRECTORY,
  // ... more config
})
```

or repeating the key in your `.env` file, or defining the environment variable on the command line, in addition to the config you wrote in your setup script. Compare this to supporting the config options out of the box:

```js
require('dotenv').config()

new S3rver({
  directory: process.env.TMP_S3_DIRECTORY,
  defaultAccessKeyId: process.env.AWS_ACCESS_KEY_ID,
  // ... more config
})
```

The way that the `bin` file is written means that all the default constructor options can be supported by the CLI without any extra code, we just have to document the options.

I spent a few hours today writing a PR inspired by this one, but which accomplishes this configuration via constructor. 
Thankfully the concept of accounts is very thin and sparsely used in the project so it wasn't too difficult. I think it would be a really nice convenience to be able to configure a default user and keypair.

> I did intentionally leave out configurability in the constructor in the original dummy account implementation because to me an S3 API with the concept of a "default account" felt like technical debt for adding proper bucket ownership later. Cross-account scenarios are fairly common, but I'm also aware in that case that you're probably a large company who has the resources to easily run those tests against the real thing anyway.

I agree with you this would incur technical debt because any internal abstraction would deviate from how AWS actually handles accounts, authentication, and ownership. IMO there's no way we could faithfully model the AWS authentication model perfectly (we'd basically be building mock IAM at that point) but it is probably sufficient and useful to users of S3rver if we provide a thin abstraction that allows them to bootstrap default accounts and even add some more. It wouldn't behave perfectly like S3, but if there is demand for it, it could form the basis for work on mocking cross-account or multiple keypair use cases in the future.

kherock:
@leontastic that's all valid, thanks for taking the time to address this! I just want to clarify that what I'm really after is a zero-config solution that works for most people. If the defaults work for most people, that's fewer people to annoy when we come up with something that messes with people using more advanced configuration. Something along the lines of
```javascript
const { accessKeyId, secretAccessKey } = require('aws-sdk').config.credentials;
defaultAccessKeyId ??= accessKeyId;
defaultSecretAccessKey ??= secretAccessKey;
```
in the constructor would solve cases like #498 without introducing new config. Constructor options are definitely the correct way to set this up, and I'd like for their default values to satisfy most cases.

leontastic:
@kherock That is a very good point, I will see if I can incorporate values from `aws-sdk` into the defaults in the refactor. I like the idea of bothering users less for their config if it's not important to them, and letting them override if they want to test something specific (e.g. auth with signing tokens https://github.com/jamhall/s3rver/issues/659 https://github.com/jamhall/s3rver/issues/534).

PR #626, opened 2020-06-04
Reimplement Object POST with busboy with broader coverage

### LINKED ISSUE #529: Add support for dispatching events with PresignedUrl POST completion
I noticed that `s3rver` won't emit an S3 event upon completion of a POST from a Presigned Url.

The relevant code to trigger an event appear to be missing from [lib/controllers/object.js#L267](https://github.com/jamhall/s3rver/blob/master/lib/controllers/object.js#L267).

Does it make sense for this to emit an event such as `s3:ObjectCreated:Post`? I've gotten a slightly functional sample locally, but it isn't PR worthy just yet. I just verified that uploading a file via POST to a Presigned Url on S3 does indeed trigger an event `s3:ObjectCreated:Post`, which triggered my lambda with the config:
```yaml
    events:
      - s3:
          bucket: my-bucket
          event: s3:ObjectCreated:*
          rules:
            - prefix: uploads/
            - suffix: .mp4
          existing: true
```

PS: Thanks for your package, it's been fantastic to use!

Original issue here: ar90n/serverless-s3-local/issues/76

### LINKED ISSUE #621: Uploading files using multipart/form-data is case-sensitive
When I run the server on port 5000, and try to upload a file to a `my_bucket` using cUrl and using upper case in the field content-type, as it is mentioned in the command line below:

`curl -v 'http://localhost:5000/bucket_name' 
  -H 'Content-Type: multipart/form-data' 
  -F key=test.pdf 
  -F AWSAccessKeyId=S3RVER 
  -F acl=public-read 
  -F Content-Type=application/pdf 
  -F file=@test.pdf
`

the file is saved with the default content-type (`binary/octet-stream`).

But when changing the field to lower case:

`curl -v 'http://localhost:5000/bucket_name' 
  -H 'Content-Type: multipart/form-data' 
  -F key=test.pdf 
  -F AWSAccessKeyId=S3RVER 
  -F acl=public-read 
  -F content-type=application/pdf 
  -F file=@test.pdf
`

the content-type of the saved file is `application/pdf`.

So, we need to only use the lowercase to specify the content type of the uploaded file.

kherock:
Fixes #621, resolves #529 (and closes #533).

This was brought on primarily because `multiparty` is currently [suffering from some Node 14 compatibility issues](https://github.com/pillarjs/multiparty/issues/226) with no fix in sight. I'm preferring `busboy` here because it's low level enough to emulate S3's quirks w.r.t. the order it parses fields while avoiding storing intermediate files on the file system.

This approach has one tradeoff: it's not possible to reject requests with more than one `file` field (or if it were to reject those requests, it would need a way to roll back changes). I think this is a fair trade of accuracy for code simplicity, at least for now. I have some plans to rectify this when versioning is implemented. In the meantime, S3rver will still log a warning:
```
WARN: A POST request for object %s in bucket %s contained %d files in the multipart body. This request would be rejected by real S3 servers.
```


kherock:
@allan-simon just thought I'd give you a heads-up that your original implementation defaulted to `201 Created` status codes with a response body. If you're relying on that behavior in your tests, you'll need to add `success_action_status=201` to your forms since S3 uses `204 No Content` by default.

allan-simon:
@kherock  thanks for the heads up, 204 is fine :)

PR #629, opened 2020-06-06
Fix error when complete multipart upload

igk1972:
Before fix - JSON parse error. Because part.Etag is string.

After fix - upload success.

### DIFF
--- a/lib/controllers/object.js
+++ b/lib/controllers/object.js
@@ -485,7 +485,7 @@ exports.completeMultipartUpload = async function completeMultipartUpload(ctx) {
     .concat(ctx.request.body.CompleteMultipartUpload.Part)
     .map(part => ({
       number: Number(part.PartNumber),
-      etag: JSON.parse(part.ETag),
+      etag: JSON.stringify(part.ETag),
     }));
   try {
     const { md5, size } = await ctx.store.putObjectMultipart(


igk1972:
mc - minio client (S3 compatible)

```
export MC_HOST_S3=http://S3RVER:S3RVER@localhost:4568

mc cp Space_IIS.hd1080.mp4 S3/files/
Space_IIS.1080.mp4:  1.20 GiB / 1.20 GiB  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  8.74 MiB/s 2m33s
```

kherock:
Hi @igk1972, thanks for bringing this to my attention. It looks like S3 loosely accepts ETags in both their quoted and unquoted forms. Currently your implementation would double-encode the MD5 hash which would mess up part verification (currently unimplemented).

S3 is actually _very_ loose with ETag verification, it seems like it will simply strip all `"`, even if they're in the middle of a string - e.g. `"""952d2c56d048"5958336747bc"""dd98590d"` will would be an acceptable ETag value for a part with contents `Hello!` .

igk1972:
> Hi @igk1972, thanks for bringing this to my attention. It looks like S3 loosely accepts ETags in both their quoted and unquoted forms. Currently your implementation would double-encode the MD5 hash which would mess up part verification (currently unimplemented).
> 
So what to do?
Remove JSON.stringify?

kherock:
I have a PR ready matching S3's behavior more accurately and included some additional test coverage for it. It'll be in the next minor release which should be out soon!

igk1972:
Thank you

Issue #567, opened 2019-11-28
HEAD got 403 (using boto python library)


allan-simon:
When I do this very simple code with version 3.5.0 I got a 403

```python
import boto                                                                                                                                                                                  
import boto.s3.connection                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                    
bucket = boto.connect_s3(                                                                                                                                                                    
    host="127.0.0.1",                                                                                                                                                                               
    port=5001,                                                                                                                                                                               
    is_secure=False,                                                                                                                                                                         
    calling_format=boto.s3.connection.OrdinaryCallingFormat(),                                                                                                                               
).get_bucket("yolaw-develop")
```

and I launch the server with 

```sh
s3rver -a 0.0.0.0 --port 5001 --directory /tmp --configure-bucket "yolaw-develop" /CORS.xml
```

I got the following log fom the server

```
info: HEAD /yolaw-develop/ 403 1ms -                                                                                                                                                         
```

I correctly set the env variable to `S3RVER`

```
$ env | grep AWS | grep ACCE

AWS_ACCESS_KEY_ID=S3RVER
AWS_SECRET_ACCESS_KEY=S3RVER
```

allan-simon:
and tcpdump of the network communication

```
HEAD /yolaw-develop/ HTTP/1.1
Host: s3:5001
Accept-Encoding: identity
Date: Thu, 28 Nov 2019 22:11:29 GMT
Content-Length: 0
Authorization: AWS S3RVER:g7Y97yEsmOLVMCenkRmaU8L4cLM=
User-Agent: Boto/2.42.0 Python/2.7.15+ Linux/5.0.0-17.1-liquorix-amd64

HTTP/1.1 403 Forbidden
Content-Type: application/xml
Content-Length: 546
Date: Thu, 28 Nov 2019 22:11:29 GMT
Connection: keep-alive
```

allan-simon:
If I do it without the `Authorization` header I got `404`

allan-simon:
if I replay with `netcat` the same request but with `GET` instead , I got 

```
GET /yolaw-develop/ HTTP/1.1
Host: s3:5001
Accept-Encoding: identity
Date: Thu, 28 Nov 2019 22:11:29 GMT
Content-Length: 0
Authorization: AWS S3RVER:g7Y97yEsmOLVMCenkRmaU8L4cLM=
[...]

PR #630, opened 2020-06-07
Fix v2 signature verification

### LINKED ISSUE #513: SignatureDoesNotMatch error when signing a request containing unicode characters

kherock:
Originally I was under the impression that our test suite ran using v2 signatures for all requests, but they're actually defaulting to v4 signatures (and S3rver makes no attempt to verify those). Turning on v2 signatures makes a whopping 101/184 tests fail!

This PR fixes all v2 verification errors within the testing suite and resolves #513 (closes #526).

### DIFF
--- a/lib/middleware/authentication.js
+++ b/lib/middleware/authentication.js
@@ -7,7 +7,11 @@ const querystring = require('querystring');
 const AWSAccount = require('../models/account');
 const S3Error = require('../models/error');
 const { RESPONSE_HEADERS } = require('./response-header-override');
-const { parseDate, parseISO8601String } = require('../utils');
+const {
+  encodeURIComponentRFC3986,
+  parseDate,
+  parseISO8601String,
+} = require('../utils');
 
 const SUBRESOURCES = {
   acl: 1,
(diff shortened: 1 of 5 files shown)


kherock:
@allan-simon v2 signature verification was actually quite broken and only functioned as expected for pre-signed URLs. I'm pretty confident #567 is fixed by this, but I haven't tested outside of the JavaScript AWS SDK.

If you have time to test against this branch, please let me know if it's still not working for you. Once v3.6.0 is published I'll assume it's fixed otherwise.

allan-simon:
@kherock  thanks for the ping :)  looking to the code I think it will fix it yes ! don't wait for me :)

Issue #627, opened 2020-06-04
Migrate to GitHub Actions


kherock:
This would replace Travis CI. I'd basically be copying this workflow verbatim:

https://github.com/webpack-contrib/css-loader/blob/master/.github/workflows/nodejs.yml

GH Actions doesn't seem to provide any totally distinct advantages over Travis, but I do love the workflow syntax of Actions. I'd also like to use this as an opportunity to actually test Windows environments since this project already adds several workarounds for those scenarios.

Issue #624, opened 2020-06-03
Publish official Docker images


kherock:
With [the deprecation of path-style URLs this year](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) I think it's important to figure out how to make developing without path-style access as frictionless as possible.

Fortunately, the style of this project very naturally lends itself to running in a container, and virtual networks make it easy to create custom DNS setups without modifying system-level DNS settings. So, I think it's time we started publishing official docker images.

kherock:
@jamhall when you have a moment, is [this](https://hub.docker.com/u/jamhall) your DockerHub page? If you're comfortable with it, I'd like to have collaborator access to `jamhall/s3rver`. For me it's important that non-Docker-official images are published under the name of the author owning the source code. In the meantime I'll plan to publish things under my forked repo's name.

jamhall:
@kherock  - great idea for having official Docker images. Yups, that is my DockerHub page.  Is [this](https://hub.docker.com/u/kherock) your DockerHub page? If so, then I will add you to s3rver [repository](https://hub.docker.com/repository/docker/jamhall/s3rver)

Thanks :)

kherock:
Yep, that's me! Thanks for setting that up.

jamhall:
Great. I've added you.

PR #637, opened 2020-06-11
Added support for Docker, GitHub Actions and graceful shutdown

ad-m:


ad-m:
Related #624 (missing setup login and secrets of registry), #627 (missing cleanup TravisCI)

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	npm run eslint
+	npm run prettier-check
+
+.PHONY: build
+build:
+	npm install
+
+.PHONY: test
+test:
+	npm run test
```
fishcharlie, line 15: Not quite sure I understand the purpose for why we need a `Makefile` for these cases. What is the benefit of running `make test` vs `npm run test`? Doesn't really seem like an advantage to me.

.github/workflows/javascript.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '${{ matrix.node }}'
+    - name: Build
+      run: make build
+    - name: Test
+      run: make test
```
fishcharlie, line 22: Wouldn't we want another job to run the `lint` as well?

fishcharlie (changes requested):

Makefile
kherock: At the risk of sounding unconstructive, I was also a little put off by the use of Makefile. I do think most of these would function fine as npm scripts as this is a relatively simple JavaScript library.

ad-m: I personally don't like have non-JavaScript things like Docker in `package.json`. `package.json` does not support variables what is useful for build image.

However, I am open to cooperation and taking into account the suggested changes. I will drop `Makefile` and create a new PR.

kherock (commented):
I'm not totally against the Makefile for things like building a Docker image, though I'm pretty comfortable typing in `docker build . -t jamhall/s3rver` myself anytime I want to build a fresh local copy.

I've opened #681, and I'd appreciate your feedback there @ad-m. If you any serious grievances with it let me know, otherwise I think I'd like to repurpose this PR as simply "Official Docker support".

Issue #638, opened 2020-06-11
Drop support for Node 8 (EOL was Dec 2019)


ad-m:
As mentioned in #623 our aim is to drop support for Node 8 as EOL.

PR #639, opened 2020-06-11
Drop fs-extra as dependency

ad-m:
Current NodeJS versions have support for ```fs.promises```. In this case, fs-extra is not necessary. See also #638 for drop support for NodeJS 8.

### DIFF
--- a/bin/s3rver.js
+++ b/bin/s3rver.js
@@ -9,7 +9,7 @@ const pkg = require('../package.json');
 const S3rver = require('..');
 
 function ensureDirectory(directory) {
-  fs.mkdir(directory, {recursive: true});
+  fs.mkdir(directory, { recursive: true });
   return directory;
 }
 
--- a/bin/s3rver.js
+++ b/bin/s3rver.js
@@ -3,13 +3,13 @@
 
 /* eslint-disable no-console */
 
-const fs = require('fs-extra');
+const fs = require('fs');
 const program = require('commander');
 const pkg = require('../package.json');
 const S3rver = require('..');
 
 function ensureDirectory(directory) {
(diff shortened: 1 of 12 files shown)


ad-m:
Anyone have an idea why they don't pass the Node 10 tests?

leontastic:
I agree with this change in principle, but I can't merge this if the tests don't pass. Looking at the failed cases it looks like there might be an issue with directory cleanup. I'm currently unable to check out the branch to try locally because it looks like it was deleted. Maybe we can revisit this in the future as part of another PR. Thank you for your effort @ad-m

ad-m:
@leontastic you can always pull old pull request via:
```bash
git clone git@github.com:jamhall/s3rver.git
cd s3rver
git fetch origin "pull/639/head:pr-639"
```

leontastic:
@ad-m Neat! I gave her a shot and the tests are still failing on Node 10. Not sure if we still want to support Node 10 but it's probably a decision for the next major release. This is a nice-to-have but it's not that urgent.

I would reopen this branch but I can't seem to since it was deleted. Maybe re-submit a PR to track progress on making the Node 10 tests pass? It's up to you if you'd like to take the initiative. Thanks again @ad-m

ad-m:
Currently, I do not have the resources to continue the work. I deleted my fork, so GitHub show branch as "unknown repository". 

If you want - go ahead and take the code.

kherock:
Btw, the issue here is simply that the `recursive` flag wasn't added to `fs.rmdir(Sync)` until Node 12:

image

If we want to keep Node 10 support, we could bring in a more specialized library like `rimraf`. My vote is currently just to drop it v10 for the next major.

PR #660, opened 2020-08-15
Implement support for signature v4

### LINKED ISSUE #659: Signature doesn't seem to be checked
Hey folks, great work with this package!

I'm probably missing something here because I clearly see past issues referencing signature mismatch errors, but in my case the signature just doesn't appear to be checked.

I'm starting the server from command line:
```
s3rver -p 8147 -d /tmp/s3rver --key ./localhost/localhost.key --cert ./localhost/localhost.crt --configure-bucket my-bucket ./ops/local-s3-cors-config.xml
```

Then from my js code I make a PUT request to store a file, and it gets correctly stored:
```
https://localhost:8147/my-bucket/4a516fad-3e42-4e2e-aa9d-adaaac93a8e3?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=S3RVER%2F20200813%2Feu-west-2%2Fs3%2Faws4_request&X-Amz-Date=20200813T094944Z&X-Amz-Expires=30&X-Amz-SignedHeaders=host&X-Amz-Signature=a7f13dbe095873e7afc897e3cdbf56b4568f1b5525bdc84e34acb26f54fc4d3e
```

Note that I'm using the "S3RVER" key and secret to generate these signatures, any other credentials result in a 403.

My observation is that:
- I can use the same URL both for PUT and for GET which should not be possible as the verb is part of the signed string
- even if I intentionally corrupt the generated signature before making the PUT request, it still goes through ok

What am I missing?

idevelop:
- Implement v4 signing algorithm, minor code changes necessary to extract relevant data from the request
- Fixed failing CORS OPTIONS request when using signed headers (I think OPTIONS should not be verified at all, since it's a different HTTP method so would definitely result in a different signature).
- Resolved small issue when calculating the canonical query string regarding params with no value (v2 does not require the = to be present, v4 does), so I moved that stripping code into the v2-specific code path
- Added unit tests for incorrect v4 signatures when using headers as well as using query params
- Confirmed locally that signed GET and PUT requests work correctly, including CORS, from my web app
- Confirmed locally that corrupting a signed URL results in a 403

Fixes #659

### DIFF
--- a/lib/middleware/authentication.js
+++ b/lib/middleware/authentication.js
@@ -1,8 +1,7 @@
 'use strict';
 
-const { createHmac } = require('crypto');
+const { createHmac, createHash } = require('crypto');
 const { mapKeys, pickBy } = require('lodash');
-const querystring = require('querystring');
 
 const AWSAccount = require('../models/account');
 const S3Error = require('../models/error');
@@ -52,6 +51,11 @@ module.exports = () =>
   async function authentication(ctx, next) {
     if (ctx.state.website) {
       // skip for static website requests
+      return next();
+    }
+
+    if (ctx.method === 'OPTIONS') {
+      // skip for CORS OPTION requests
       return next();
     }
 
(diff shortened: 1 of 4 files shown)


idevelop:
@kherock is this something you'd be interested in merging?

test/middleware/authentication.spec.js
```diff
@@ -386,4 +386,44 @@ describe('REST Authentication', () => {
...
+          'X-Amz-SignedHeaders': 'host',
+          'X-Amz-Signature': 'badsig',
+        },
+      });
+    } catch (err) {
+      res = err.response;
+    }
+    expect(res.statusCode).to.equal(403);
+    expect(res.body).to.contain('<Code>SignatureDoesNotMatch</Code>');
+  });
```
fishcharlie, line 428: Seems like we should also have a test to verify that a successful signature actually works, right?

lib/middleware/authentication.js
```diff
@@ -464,35 +497,105 @@ function parseQueryV4(query) {
...
+ * Performs the calculation of an authentication code for a string using the specified key and
+ * various components required by v4.
+ *
+ * @param {String} stringToSign the string representation of a canonical request
+ * @param {String} secretKey a secret access key
+ * @param {String} date received from the credential header
+ * @param {String} region received From the credential header
+ * @param {String} service received From the credential header
+ */
+function calculateSignatureV4(stringToSign, secretKey, date, region, service) {
```
fishcharlie, line 581: Do we want to break out the V2 and V4 methods into 2 different files? That way it reduces the file size of this file, and we can separate out concerns a little bit?

fishcharlie (changes requested):

kherock:
These just needs a couple more tests but otherwise this would be a great addition.

I did want to suggest that AWS signing is a problem that has already been solved and it might be a better use of this project's time to offload request signing to another library such as [aws4](https://www.npmjs.com/package/aws4).

kherock:
I've switched the default S3 client used in tests to use v4 signing. There were some slight tweaks I needed to make for computing the canonicalized query string for both versions. Thanks for finding the existing bug in the V2 signing as well!

PR #677, opened 2021-02-11
Added support for Docker, GitHub Actions and graceful shutdown 

ad-m:
Close #637

Follow-up #637 due deleted fork


kherock (commented):
Hey, sorry I missed this before leaving a comment on your old PR. Also don't feel offended by me opening #681, those were honestly changes I've had in my own personal fork since June.

My comment [here](https://github.com/jamhall/s3rver/pull/637#discussion_r574854455) still stands. I'm not particularly attached to `4568` as the default port, it's just what we've always had. Regardless, I really think the Docker image should be consistent with the documented default port.

README.md
```diff
@@ -3,12 +3,13 @@
...
 
 [![Build Status](https://api.travis-ci.org/jamhall/s3rver.png)](https://travis-ci.org/jamhall/s3rver)
+![JavaScript files](https://github.com/jamhall/s3rver/workflows/JavaScript%20files/badge.svg)
 [![Dependency Status](https://david-dm.org/jamhall/s3rver/status.svg)](https://david-dm.org/jamhall/s3rver)
 [![Devdependency Status](https://david-dm.org/jamhall/s3rver/dev-status.svg)](https://david-dm.org/jamhall/s3rver?type=dev)
 
 S3rver is a lightweight server that responds to **some** of the same calls [Amazon S3](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html) responds to. It is extremely useful for testing S3 in a sandbox environment without actually making calls to Amazon.
 
-The goal of S3rver is to minimise runtime dependencies and be more of a development tool to test S3 calls in your code rather than a production server looking to duplicate S3 functionality.
+The goal of S3rver is to minimize runtime dependencies and be more of a development tool to test S3 calls in your code rather than a production server looking to duplicate S3 functionality.
```
jamhall, line 12: This isn't a typo or a spelling mistake. I'm British, so I use the British English spelling of words! 😃

bin/s3rver.js
```diff
@@ -91,7 +91,20 @@ program.on('--help', () => {
 program.action(async command => {
   const { configureBucket, ...opts } = command.opts();
   opts.configureBuckets = configureBucket;
-  const { address, port } = await new S3rver(opts).run();
+
+  const server = new S3rver(opts);
+  const { address, port } = await server.run();
+
+  process.on('SIGINT', async function onSigint() {
```
leontastic, line 98: From [Node docs](https://nodejs.org/api/process.html#process_signal_events):

> 'SIGTERM' and 'SIGINT' have default handlers on non-Windows platforms that reset the terminal mode before exiting with code 128 + signal number. **If one of these signals has a listener installed, its default behavior will be removed (Node.js will no longer exit).**

I think this means you must call `process.exit` inside these handlers or else the `node` process won't be terminated via `SIGINT` or `SIGTERM`, you would have to send a `SIGKILL` to terminate the process.

lib/s3rver.js
```diff
@@ -145,6 +161,8 @@ class S3rver extends Koa {
 
       const { address, port, ...listenOptions } = this.serverOptions;
       this.httpServer = await this.listen(port, address, listenOptions);
+      stoppable(this.httpServer);
```
leontastic, line 164: I think I understand what `.stop` patched by `stoppable` is supposed to do, but I'm having trouble understanding why S3rver would benefit from it compared to using the `.close` function.

S3rver is intended for integration testing, so when you call `.close` you would probably want to perform cleanup as quickly as possible and do not care about open connections. If there are in-flight requests or Keep-Alive connections still open when you call `.close`, you probably forgot to wait for your requests to end before cleaning up your tests, and you would want to receive errors about prematurely closed connections rather than have these connections silently cleaned up on close.

If you are running the process independently in the background or in a container, prematurely closed connections would indicate that you closed your server earlier than your tests completed, which IMO is fine and should be expected. It would be a little unexpected for that server to remain alive while my tests finish up their current requests, only to reject future requests and fail the remainder of the tests.

So I think there are no use cases where `.stop`'s graceful functionality would be useful, and would argue against introducing `stoppable` for this reason. However if you can suggest a use case where graceful shutdown is needed I may reverse my opinion on this.

leontastic (changes requested):
I'm 100% fine with the Docker stuff, I would approve a PR containing only those changes. However I think GitHub actions were handled in #681, and I am unconvinced about the need for graceful shutdown.

@ad-m Thanks for the initiative you took on this PR. If you can address my concerns in the review I'd be happy to come back to review this PR again.

PR #600, opened 2020-03-10
Add test coverage reporting

kkoskelin:
As an aid to ensuring proper coverage, I've installed istanbul and configured coverage numbers that ought to ensure the threshold reflects current coverage and do not slip below those given numbers.
```
------------------------------|---------|----------|---------|---------|----------------------------------------------------------------------------
File                          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                                          
------------------------------|---------|----------|---------|---------|----------------------------------------------------------------------------
All files                     |   88.38 |    76.48 |   90.87 |   88.78 |                                                                            
 lib                          |   84.76 |     51.3 |   93.33 |   84.23 |                                                                            
  routes.js                   |   78.05 |    34.83 |     100 |   77.78 | 24,38,74,79,107,112,122,151,156,168,170,184,189,200,202,213,222,227        
  s3rver.js                   |   82.64 |    65.96 |   88.24 |   82.35 | 70-72,99,111,116,122,152,153,175,191-194,196,233,237,247,253,259,260       
  utils.js                    |   96.97 |    94.44 |   93.75 |   96.67 | 65,85                                                                      
 lib/controllers              |   90.27 |    86.51 |     100 |   90.38 |                                                                            
  bucket.js                   |   90.43 |    87.84 |     100 |   91.15 | 146,157,171,184,199,206,285,290,376,380                                    
  object.js                   |   89.83 |    84.62 |     100 |   89.53 | 24,119,125,169,170,174,175,274,308,313,347,353,467,473,518,524,579,586     
  service.js                  |     100 |      100 |     100 |     100 |                                                                            
 lib/middleware               |   89.46 |    84.44 |   94.29 |   89.57 |                                                                            
  authentication.js           |    87.5 |    80.21 |     100 |   87.41 | 145,146,152,178,219,239,252,262,296,312,356,379,412,420,421,427,446        
  cors.js                     |   94.44 |    90.63 |     100 |   94.29 | 22,29                                                                      
  logger.js                   |     100 |      100 |     100 |     100 |                                                                            
  response-header-override.js |   95.65 |    92.31 |     100 |   95.45 | 35                                                                         
  vhost.js                    |     100 |      100 |     100 |     100 |                                                                            
  website.js                  |   86.24 |    82.35 |      75 |   86.79 | 16,24,49,50,174,178,188,194,200,201,239,242,247,249                        
 lib/models                   |   85.35 |    83.33 |   77.78 |   85.94 |                                                                            
[...]

PR #681, opened 2021-02-12
Implement GitHub Actions w/ codecov reports

### LINKED ISSUE #627: Migrate to GitHub Actions

kherock:
Closes #627. This includes #600 and includes the test matrix I was working on before #637 was opened. It seems like there are some open file handle issues in the POST Object middleware that's causing a decent chunk of the test suite to fail on Windows.


kherock:
@jamhall any objection with migrating to `main` as the default branch? Just trying to avoid introducing a dependency on `master` as the default branch here.

kherock:
Also, this closes #636 since `package-lock.json` is needed to run the `npm ci` and `npm audit` commands.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,75 @@
+name: s3rver
```
fishcharlie, line 1: Not a huge deal. But I tend to suggest that naming is more specific for GitHub Actions workflows. If you ever want to add workflows to close stale issues, tag PRs, publish new versions, etc. those will tend to be different workflows. Which would make this name confusing since it's so general.

For [Dynamoose](https://github.com/dynamoose/dynamoose/blob/master/.github/workflows/ci.yml#L1) we use `CI`. Not sure that makes the most amount of sense tho.

```suggestion
name: CI
```

---

I also think it'd be valuable to have the name match the file name.

---

Super nit-picky, but I thought it was worth at least mentioning.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,75 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - next
+  pull_request:
+    branches:
+      - master
+      - next
```
fishcharlie, line 11: Any purpose for limiting branches here? Are there certain branches we don't want this to run on, and why?

fishcharlie (approved):
Looks really great!!!

fishcharlie:
Is this going to be the prioritized PR or is #677?

fishcharlie:
Looks like we also have some CI failures here. Looks like most of it is around Windows FS permissions.

.github/workflows/nodejs.yml
ad-m: The branch limitation avoids double building in case of pushing and immediately creates a pull request by the project developer. However, for this purpose it is enough to limit the branches for pushs.

.github/workflows/nodejs.yml
kherock: CI makes enough sense to me, and thank you for giving me a new pet peeve for when I'm looking at other projects' actions 😄.

kherock:
> Is this going to be the prioritized PR or is #677?

There's still a decent amount of work ahead before it makes sense to start publishing Docker images for 4.x, so I'd like to get this in first.

jamhall:
@kherock 
> @jamhall any objection with migrating to `main` as the default branch? Just trying to avoid introducing a dependency on `master` as the default branch here.

Nopes. No objection from me :)

kherock:
Great, could you rename the branch from the tab in repo settings? It'll automatically move all the PRs over once you do that.

jamhall:
> Great, could you rename the branch from the tab in repo settings? It'll automatically move all the PRs over once you do that.

Hi @kherock  - it's done

PR #687, opened 2021-02-14
Fixing vhostBuckets documentation

### LINKED ISSUE #685: vhostBuckets Documentation Confusing
In the README it says the `vhostBuckets` property means `Disable vhost-style access for all buckets`.

The description makes it sound like if this property is set to `true` that it will disable vhost-style access for all buckets. However, when reading the code `vhostBuckets: true` that makes it sound like it will enable vhost style access.

Which is it?

fishcharlie:
Closes #685


kherock:
Thanks!

PR #721, opened 2021-04-23
feat(object): add support for aws-chunked uploads

### LINKED ISSUE #534: Support Signature Version 4 streaming API

kherock:
Closes #534. Using [this](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html) as my primary implementation guide in addition to inspecting [how the Java SDK implements chunk signatures](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/auth/AwsChunkedEncodingInputStream.java), I've successfully come up with a small suite of tests that pass against real S3 services.


kherock:
This ended up needing far more research than I anticipated as there aren't any JS/Node S3 clients that actually implement chunked transfer encoding. I also found a couple places in the file system store where a request stream error wasn't being handled and could lead to Promises that would never be resolved. This also sets up most of the groundwork needed to implement a Transform stream that computes payload hashes/signatures for aws4-signed requests.

PR #753, opened 2021-06-14
fix(object): Takes into account the range of bytes starting with 0

nicoandra:
Fixes returning the `Range-Content` header when requesting a range starting with byte 0 and no end, ie `bytes=0-`

The issue has been reported here: https://github.com/jamhall/s3rver/issues/754


nicoandra:
Hey @jamhall , added tests that cover the specific case I've experienced when using `smart-open` with this library.

kherock:
Closing in favor of #755

Issue #754, opened 2021-06-14
Content Range header not returned when requested range is 'bytes=0-'


nicoandra:
Had some trouble using this library with [smart_open](https://pypi.org/project/smart-open/)

See the snippet below, where I hit both a real S3 implementation and the S3rver implementation:

```
## Install dependencies:
pip install smart-open
```

Code
```
from boto3.session import Session
from smart_open import open

def create_s3_client_real(): 
    client = Session(
        aws_access_key_id="KEY", 
        aws_secret_access_key="SECRET"
    ).client( "s3", )
    return client

def create_s3_client_s3rver(): 
    client = Session(
        aws_access_key_id="S3RVER", 
        aws_secret_access_key="S3RVER"
    ).client("s3", endpoint_url="http://localhost:4569")
    return client

real = create_s3_client_real()
s3rver = create_s3_client_s3rver()

print(real.get_object(Bucket="bucket", Key="object.txt", Range="bytes=0-")) 
print(" *********************** ") 
print(s3rver.get_object(Bucket="bucket", Key="object.txt", Range="bytes=0-"))

```

The problem arises because the check `if(options.start || options.end)` evaluates to false when `options.start = 0`; however it is a valid range. 

A Pull Request has been submitted: https://github.com/jamhall/s3rver/pull/753

kherock:
Closed via #755 and released in v3.7.1!

PR #755, opened 2021-06-15
fix(object): Takes into account the range of bytes starting with 0

nicoandra:
Fixes errors when trying to re-create directories that already exist.

Uses a helper function to verify for the directory existence first.

Note this PR also includes the fix for `options.start !== undefined` as I've built on top of my previous branch.

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -16,7 +16,7 @@ const { format } = require('util');
 const { getConfigModel } = require('../models/config');
 const S3Bucket = require('../models/bucket');
 const S3Object = require('../models/object');
-const { concatStreams, walk } = require('../utils');
+const { concatStreams, walk, ensureDir } = require('../utils');
 
 const S3RVER_SUFFIX = '%s._S3rver_%s';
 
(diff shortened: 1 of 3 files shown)


lib/stores/filesystem.js
```diff
@@ -48,6 +48,19 @@ class FilesystemStore {
     return path.join(this.rootDirectory, bucketName);
   }
 
+  async makeBucketDirIfNotExists(bucketPath) {
+    let exists = false;
+    try {
+      await fs.access(bucketPath, fs.constants.F_OK);
+    } catch (ex) {
+      exists = true;
```
kherock, line 56: Thanks for refactoring this, I wasn't sure exactly where the Windows tests were going wrong since I replaced fs-extra.

Could you move this to the utils file as `ensureDir(path)`? You should also check that the exception code is `ENOENT` and rethrow when it's not. You can also move the `mkdir` call below directly to this catch block instead of storing a temporary variable.

kherock (changes requested):
These look pretty good to me, do you think you could add a quick test case with the range that wasn't working before? There should be some existing tests that you can borrow from.

nicoandra:
> These look pretty good to me, do you think you could add a quick test case with the range that wasn't working before? There should be some existing tests that you can borrow from.

Hey @kherock , the ranges changes were done in a separate PR: https://github.com/jamhall/s3rver/pull/753

I'll revisit this PR again to refactor a few more things and get back to you. Thanks!

kherock (approved):
This looks good, thanks. I'll merge and backport to 3.x when I have time.

non-senses:
This PR is related to this issue: https://github.com/jamhall/s3rver/issues/754

PR #777, opened 2021-09-24
docs: option address, not hostname

jasonprado:
The field is `address` in the code, not `hostname`. Setting `hostname` has no effect.

See: https://github.com/jamhall/s3rver/blob/main/lib/s3rver.js#L146


kherock:
Thanks @jasonprado!

PR #778, opened 2021-10-02
Remove duplicate code for save content in store

ad-m:
I have notice some code duplicated, so I suggest extract helper method in store for that.

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -106,6 +106,35 @@ class FilesystemStore {
     await fs.writeFile(metadataPath, JSON.stringify(json, null, 2));
   }
 
+  async putContent(objectPath, content) {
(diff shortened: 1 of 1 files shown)


ad-m:
Taking into account previous runs eg. https://github.com/jamhall/s3rver/runs/3707557361?check_suite_focus=true that failure looks as tests flakiness.

kherock:
Hey! Thanks for this, there was another PR I just merged that fixes the test flakiness on Windows. It's a regression from dropping fs-extra a while back. It should pass once you rebase your branch.

ad-m:
The PR has been rebased. 

It seems to me that the decrase in test coverage is not in line with this change, which limits code inflation.

kherock:
Yeah no worries about the coverage, thanks @ad-m!

PR #784, opened 2021-10-23
add operation to abort multipart upload

ad-m:
I have noted missing operation to abort multipart upload. That operation does not have response, so it's very easy to implement.

### DIFF
--- a/lib/controllers/object.js
+++ b/lib/controllers/object.js
@@ -803,3 +803,20 @@ exports.uploadPartCopy = async function uploadPartCopy(ctx) {
     throw err;
   }
 };
+
+/**
+ * Abort Multipart Upload
+ * This action aborts a multipart upload. After a multipart upload is aborted,
+ * no additional parts can be uploaded using that upload ID. The storage
+ * consumed by any previously uploaded parts will be freed.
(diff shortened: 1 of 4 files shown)


lib/controllers/object.js
```diff
@@ -803,3 +803,20 @@ exports.uploadPartCopy = async function uploadPartCopy(ctx) {
...
+ * no additional parts can be uploaded using that upload ID. The storage
+ * consumed by any previously uploaded parts will be freed.
+ * {@link https://docs.aws.amazon.com/AmazonS3/latest/API/API_AbortMultipartUpload.html}
+ */
+exports.abortMultipartUpload = async function abortMultipartUpload(ctx) {
+  await ctx.store.abortUpload(
+    ctx.params.bucket,
+    ctx.params.key,
+    ctx.query.uploadId,
+    ctx.headers,
```
ferhatelmas, line 819: why to pass unused (key & headers) ?

lib/controllers/object.js
```diff
@@ -803,3 +803,20 @@ exports.uploadPartCopy = async function uploadPartCopy(ctx) {
...
+  );
+  ctx.body = '';
```
ferhatelmas, line 821: will need to set status too (204)

ferhatelmas:
@jamhall is it possible to make a release with this ?

PR #789, opened 2022-01-03
Fix race condition when deleting files

### LINKED ISSUE #788: Apparent race condition deleting objects results in ENOENT
If I issue concurrent delete requests for objects that share a path, then the operation fails with a 500. Here's a sample typescript test using jest and the aws-sdk client library.

I think the root cause is that both are trying to do an rmdir up the chain of directories but they end up conflicting. Maybe `rmdir` should swallow ENOENT exceptions?

```
import { AWSError, S3 } from 'aws-sdk'
import S3rver from 's3rver'
import * as tmp from 'tmp'

function sync<T>(r: AWS.Request<T, AWS.AWSError>): Promise<T> {
  return new Promise((resolve, reject) => {
    r.send((err, data) => {
      if (err) {
        reject(err)
      } else {
        resolve(data)
      }
    })
  })
}

let s3: S3
let server: S3rver

beforeAll(() => {
    const serverDir = tmp.dirSync({ prefix: 's3-api.test', unsafeCleanup: true })

    server = new S3rver({
      port: 0,
      silent: false,
      directory: serverDir.name,
      configureBuckets: [
        {
          name: 'my-bucket',
          configs: [],
        },
      ],
    })
    const bound = await server.run()

    const s3 = new S3({
      region: 'us-east-1',
      accessKeyId: 'S3RVER',
      secretAccessKey: 'S3RVER',
      endpoint: `http://${bound.address}:${bound.port}`,
      s3ForcePathStyle: true,
    })
}

afterAll(async () => {
      await server.close()
})

test('s3rver racing delete', async ()  => {
    await Promise.all([
      sync(
        s3.putObject({
          Bucket: 'my-bucket',
          Key: '8ccf8a22d464800dccc55b90791fc6c43a867aeebda6d2772839f8a2105f9977/71f3db21645ab17a0f6fb561dad81608521e80a34d178646f738a98cb2ae8de4/0aa7509751f6c461ccd91c19e90360cccae3889ca83ecaa0415ff3bc0260f161/93bb309257da197aded081fa3de9c2c0',
        }),
      ),
      sync(
        s3.putObject({
          Bucket: 'my-bucket',
          Key: '8ccf8a22d464800dccc55b90791fc6c43a867aeebda6d2772839f8a2105f9977/71f3db21645ab17a0f6fb561dad81608521e80a34d178646f738a98cb2ae8de4/0aa7509751f6c461ccd91c19e90360cccae3889ca83ecaa0415ff3bc0260f161/93bb309257da197aded081fa3de9c2c0',
        }),
      ),
    ])

    await Promise.all([
      sync(
        s3.deleteObject({
          Bucket: 'my-bucket',
          Key: '8ccf8a22d464800dccc55b90791fc6c43a867aeebda6d2772839f8a2105f9977/71f3db21645ab17a0f6fb561dad81608521e80a34d178646f738a98cb2ae8de4/0aa7509751f6c461ccd91c19e90360cccae3889ca83ecaa0415ff3bc0260f161/93bb309257da197aded081fa3de9c2c0',
        }),
      ),
      sync(
        s3.deleteObject({
          Bucket: 'my-bucket',
          Key: '8ccf8a22d464800dccc55b90791fc6c43a867aeebda6d2772839f8a2105f9977/71f3db21645ab17a0f6fb561dad81608521e80a34d178646f738a98cb2ae8de4/0aa7509751f6c461ccd91c19e90360cccae3889ca83ecaa0415ff3bc0260f161/93bb309257da197aded081fa3de9c2c0',
        }),
      ),
    ])
}

### PR DESCRIPTION
**Pull request by vlovich**
Gracefully handle multiple threads deleting an object at the same time.
When two concurrent requests come in to delete the last two remaining
objects in a given path, they'll be racing each other trying to delete
and cause a spurious deletion failure for one of them.

Fixes #788

### DIFF
--- a/lib/stores/filesystem.js
+++ b/lib/stores/filesystem.js
@@ -377,14 +377,14 @@ class FilesystemStore {
     const parts = key.split('/');
     // the last part isn't a directory (it's embedded into the file name)
     parts.pop();
-    while (
-      parts.length &&
-      !readdirSync(path.join(bucketPath, ...parts)).length
-    ) {
-      await fs.rmdir(path.join(bucketPath, ...parts));
+    while (parts.length) {
+      await fs.rmdir(path.join(bucketPath, ...parts)).catch(err => {
+        if (err.code !== 'ENOENT' && err.code !== 'ENOTEMPTY') throw err;
+        parts.length = 0;
+      });
       parts.pop();
     }
-  }
+}
 
   async initiateUpload(bucket, key, uploadId, metadata) {
     const uploadDir = path.join(

### CODE REVIEW

File: lib/stores/filesystem.js [hunk 1]
```diff
@@ -377,14 +377,13 @@ class FilesystemStore {
     const parts = key.split('/');
     // the last part isn't a directory (it's embedded into the file name)
     parts.pop();
-    while (
-      parts.length &&
-      !readdirSync(path.join(bucketPath, ...parts)).length
-    ) {
-      await fs.rmdir(path.join(bucketPath, ...parts));
-      parts.pop();
```
leontastic, line 385: I think you forgot to re-add this line. Otherwise empty parent directories don't get cleaned up. You actually included it in your example here: https://github.com/jamhall/s3rver/issues/788#issuecomment-995408847

leontastic (commented):
@vlovich Thank you for your PR contribution.

I read [the issue you filed](https://github.com/jamhall/s3rver/issues/788) and agree with your reasoning on the solution. I think you're missing a `parts.pop()` call, see my comment in my review.

Give it a test and let me know. I haven't touched `npm` in years but I can probably publish a release for this bugfix after it's merged.

lib/stores/filesystem.js
vlovich: Sorry for the delayed reply. I hadn't seen this. Good catch. I left the pop in there locally but somehow corrupted the patch when applying.

PR #803, opened 2022-08-15
Add support for ${filename} in POST Object

RichardBradley:
The S3 POST Object api allows Key to contain "`${filename}`" in order to use the client-provided filename to form part of the uploaded Object Key. (Notably, no escape mechanism is provided if you in fact wanted a Key containing `${filename}`.)

See https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html

This PR adds support for this feature in `s3rver`

Thanks for a great tool!

Rich


RichardBradley:
Any update on this?

We have been using this in our test harnesses for a while and it works fine for us

